PR #4, opened 2019-01-27
static: add redirect for go-llvm

aykevl:
go-llvm has moved, add a redirect for vanity imports.


deadprogram:
Looks good. Now merging.

PR #5, opened 2019-01-29
Language support in docs

aykevl:
Add a page that describes which part of Go is supported and which is not. Most features require some explanation as they are only partially supported.

Also, some updates to the "heap allocation" page.

When this is live, I would like to replace the part in the README with a link to this page.


deadprogram:
Seems like some good and needed info. Now merging/deploying.

PR #9, opened 2019-02-15
fix wasm docs example

didil:


deadprogram:
Now merging, thank you @didil

PR #10, opened 2019-03-12
Corrected path to include "src"

DazWilkin:


deadprogram:
Thanks @DazWilkin for the correction. Now merging.

PR #11, opened 2019-03-17
Augment OpenOCD requirement with file copy

DazWilkin:
Rather than install `OpenOCD`, Micro:bit devices mount as USB devices ([DAPLink](https://github.com/ARMmbed/DAPLink))when connected to a host and it is possible to flash software to them using a simply copy command.

This *may* apply to other supported devices too and provides an easier path to flashing.

### DIFF
--- a/content/microcontrollers/bbc-microbit.md
+++ b/content/microcontrollers/bbc-microbit.md
@@ -3,7 +3,7 @@ title: "BBC:Microbit"
 weight: 3
 ---
 
-The [BBC micro:bit](https://microbit.org) is a tiny programmable computer designed for learning. It is based on the Nordic Semiconductor [nRF51822](https://www.nordicsemi.com/eng/Products/Bluetooth-low-energy/nRF51822) ARM Cortex MO chip.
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @DazWilkin I think you have the right idea here overall. 

Perhaps you can mimic the format used by https://github.com/tinygo-org/tinygo-site/blob/master/content/microcontrollers/circuit-playground-express.md#uf2 as more complete example?

deadprogram:
I was expecting something like this:

The microbit comes with the [DAPLink bootloader](#link) already installed.

- Plug your microbit into your computer's USB port.
- The microbit board will appear to your computer like a USB drive. Determine the path to the board, for example on Linux it will be something like `/media/[USERNAME]/[NAME OF THE BOARD]`.
- Build your TinyGo program to the board in `.hex` format using the `tinygo build -o=/media/[USERNAME]/[NAME OF THE BOARD]/flash.hex -target=microbit [PATH TO YOUR PROGRAM]` command.
- The microbit board should restart and then begin running your program.

deadprogram:
Thanks for the docs and making the requested changes. Now merging.

PR #16, opened 2019-04-18
drivers: add page with links to TinyGo drivers repo, and list of drivers

deadprogram:
This PR adds a page with links to TinyGo drivers repo and list of supported drivers.

### DIFF
--- a/content/microcontrollers/drivers/_index.md
+++ b/content/microcontrollers/drivers/_index.md
@@ -1,0 +1,30 @@
+---
+title: "Drivers"
+chapter: true
+weight: 90
+---
+
+# Drivers
+
+TinyGo has driver support for 14 different sensors and other devices, such as digital accelerometers and multicolor LEDs.
+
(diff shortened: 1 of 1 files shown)


aykevl:
Should we really include a list of drivers in the docs? I'm afraid it will turn stale very quickly.
We could also just provide a link to the drivers repo or the godoc list of drivers.

deadprogram:
There are a couple of good reasons to add on the site itself. One is that i think few people are able to find this list currently. That is indicated in part by the large difference in Github stars between the main repo, and the drivers repo. Another is for search engine optimization.

aykevl:
Okay. Perhaps it would be useful to add something like this? I'm afraid people will look at this list, conclude that a certain device is unsupported, and go elsewhere (or even duplicate work by writing their own).

> More drivers get added all the time. If your device isn't listed here, check the drivers repository.

deadprogram:
Added some explanatory text/link pointing user to drivers repo.

PR #19, opened 2019-04-23
getting-started: clarify and extend "install from source" instructions

aykevl:
Installing TinyGo using system libraries and by installing LLVM from
source is done very differently. Add information about using system
libraries and make the distinction clear.

This should avoid confusion like in https://github.com/tinygo-org/tinygo/issues/297.
Please note that I haven't managed to get Hugo to work, it somehow ignores the new page I've added. So it should be checked before merging this PR.

### DIFF
--- a/content/getting-started/_index.md
+++ b/content/getting-started/_index.md
@@ -6,4 +6,4 @@ weight: 1
 
 # Getting Started
 
-To use TinyGo, you must install it on your system. We have instructions for Linux, macOS, Windows, and Docker.
+To use TinyGo, you must install it on your system. We have instructions for Linux, macOS, Windows, and Docker. You can also install TinyGo from source.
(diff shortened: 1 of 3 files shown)


deadprogram:
Well, I really did not want to combine the instructions for a single OS together. Too many people are confused by this, hence why the install from source was separated by host OS. I learned this the hard way on other projects.

aykevl:
I could split it up, but it's a lot of duplication. Basically the only difference is how to install dependencies, and users familiar with a certain OS will know that you cannot run `apt-get` on macOS (or vica versa). So will people really be confused by this?

I tried to make it very clear which instructions are for which OS.

deadprogram:
I prefer some duplication and minimize the required knowledge for people trying to install from source, personally. Some of that can later be reduced, by further improving the automation of the install from source make tasks.

aykevl:
Ok, changes made.

deadprogram:
Thanks. Now merging.

PR #20, opened 2019-04-25
getting-started: add arch linux instructions

johanbrandhorst:
Tested on my own machine

### DIFF
--- a/content/getting-started/linux.md
+++ b/content/getting-started/linux.md
@@ -122,6 +122,11 @@ sudo apt-get update
 sudo apt-get install clang-8 llvm-8-dev lld-8 libclang-8-dev
 ```
 
+For Arch Linux, install the [clang](https://www.archlinux.org/packages/extra/x86_64/clang/),
+[llvm](https://www.archlinux.org/packages/extra/x86_64/llvm/) and
+[lld](https://www.archlinux.org/packages/extra/x86_64/lld/) packages.
+They must be version 8.x.
+
 Installing TinyGo should now be as easy as:
 
 ```shell


content/getting-started/linux.md
```diff
@@ -122,6 +122,9 @@ sudo apt-get update
 sudo apt-get install clang-8 llvm-8-dev lld-8 libclang-8-dev
 ```
 
+For Arch Linux, install [the llvm package](https://www.archlinux.org/packages/extra/x86_64/llvm/).
+It must be version 8.0.0 or above.
```
aykevl: > It must be version 8.0.0 or above.

But it must be LLVM 8 (8.1.0 would be OK, if that gets released).

johanbrandhorst: I wanted to illustrate anything => 8.0.0, is the strict requirement  8.0.0 <= x < 9.0.0?

johanbrandhorst:
Wait, stop, dont' merge this, the archi linux package is missing wasm-ld.

johanbrandhorst:
I'll do some more investigating

johanbrandhorst:
OK so we also need the `lld` package, but disappointingly it seems `tinygo` looks for `wasm-ld-8` and can't find `wasm-ld`. I'll raise a separate issue about that.

johanbrandhorst:
Raise https://github.com/tinygo-org/tinygo/issues/305, but the solution is not obvious to me yet. Feedback appreciated. These instructions can go in as they are now that I've added the symlink instructions.

johanbrandhorst:
I've updated this based on https://github.com/tinygo-org/tinygo/pull/323 being merged. Do you have anymore comments @aykevl?

content/getting-started/linux.md
aykevl: Oops, never replied to this.
Yes, the current requirement is on 8.x.x, so `8.0.0 <= x < 9.0.0`.

aykevl:
> Do you have anymore comments

Yes, see above.

content/getting-started/linux.md
johanbrandhorst: Updated

content/getting-started/linux.md
```diff
@@ -122,6 +122,10 @@ sudo apt-get update
 sudo apt-get install clang-8 llvm-8-dev lld-8 libclang-8-dev
 ```
 
+For Arch Linux, install the [llvm](https://www.archlinux.org/packages/extra/x86_64/llvm/)
+and [ldd](https://www.archlinux.org/packages/extra/x86_64/ldd/) packages.
```
aykevl: Reading it again, you only mention llvm and lld. Clang is also a dependency, or will it be installed along with one of those packages?

johanbrandhorst: Great question, I thought it was a dependency but I might just list it explicitly. Will update!

johanbrandhorst: Updated

aykevl:
Thank you!

Issue #23, opened 2019-05-09
extending PATH if llvm is installed using homebrew on macos


supcik:
On macos, when llvm is installed using homebrew, there is no automatic symlink in /usr/local/bin. This message is displayed after the installation on llvm:

```
llvm is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.
```

If the user did not update the PATH, tinygo might not find the llvm binaries.

I solved it by extending the PATH while running the command:

```
PATH=/usr/local/Cellar/llvm/8.0.0/bin:$PATH tinygo ...
```

justinclift:
Good point, the docs should mention this. :smile:

Probably need to add matching CFLAGS/CXXFLAGS/LDFLAGS environment variables as well, so the needed llvm includes and libraries can be found too. :smile:

As a thought for the correct path, `/usr/local/Cellar/anything` is probably not the right thing.

Is there some kind of `llvm` entry under your `/usr/local/opt`?

supcik:
You are right concerning the correct path and I have indeed symlinks in /usr/opt/local:

```
/usr/local/opt/llvm -> ../Cellar/llvm/8.0.0
/usr/local/opt/llvm@8 -> ../Cellar/llvm/8.0.0
```

This is the full info displayed after the installation of llvm:

```
To use the bundled libc++ please add the following LDFLAGS:
  LDFLAGS="-L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib"

llvm is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have llvm first in your PATH run:
  echo 'export PATH="/usr/local/opt/llvm/bin:$PATH"' >> ~/.zshrc

For compilers to find llvm you may need to set:
  export LDFLAGS="-L/usr/local/opt/llvm/lib"
  export CPPFLAGS="-I/usr/local/opt/llvm/include"
```

justinclift:
Excellent.  Yep, that `PATH` statement will be the correct one to use. :smile:

Fairly sure the `LDFLAGS` and `CPPFLAGS` ones at the end there will be needed too.

Might need to append them with themselves, so any existing `LDFLAGS` and or `CPPFLAGS` aren't overwritten:

```
export LDFLAGS="-L/usr/local/opt/llvm/lib:$LDFLAGS"
[...]

PR #25, opened 2019-05-09
add info concerning llvm installed using homebrew

supcik:
see issue #23


justinclift:
Excellent. :smile:

Looks good.  Let's figure out the best path to use (as per #23), and also whether we need to add the CFLAGS/CXXFLAGS/LDFLAGS environment variables.

justinclift:
This can probably be closed, as the correct path for Homebrew is hard coded these days. :smile:

Issue #24, opened 2019-05-09
cannot find package "unsafe" and defining GOROOT


supcik:
I installed go using homebrew on macos and I currently have the version 1.12.5.

When I run tinygo, I got the following error:

```
error: cannot find package "unsafe" in any of:
   /usr/local/Cellar/go/1.11.1/libexec/src/unsafe (from $GOROOT)
   /<HOME>/go/src/unsafe (from $GOPATH)
```

I don't have any GOROOT defined in my environment.

I was able to solve this issue by defining GOROOT while using tinygo:

```
GOROOT=$(go env GOROOT) tinygo ...
```

As I also had to extend PATH, here is my final workaround:

```
PATH=/usr/local/Cellar/llvm/8.0.0/bin:$PATH GOROOT=$(go env GOROOT) tinygo ...
```

N.B. Using `$(go env GOROOT)` works in zsh and in bash, but it might not work with all shells (e.g fish).

deadprogram:
Hi @supcik that was corrected by https://github.com/tinygo-org/tinygo/pull/330 which will be part of the next TinyGo release. Thanks!

supcik:
OK. Thank you for the update. So I guess that you can ignore the PR.

justinclift:
Depends which goes live first, the FAQ entry or the next TinyGo release. :smile:

deadprogram:
The fix in TinyGo itself was merged a while ago, but this PR was not closed in a timely way, sorry about that. Now closing, thanks again for the help!

PR #26, opened 2019-05-09
add faq for error Cannot find package "unsafe"

supcik:
see issue #24


justinclift:
Sounds good. :smile:

The `PATH` fragment at the end will probably need adjusting as per #23.

This `unsafe` error should go away with the next release too AFAIK.  The problem (in theory) should be fixed in the current development branch with 141a70f4015c0a40ccd2ade8062ff1f5214a6aea.

PR #32, opened 2019-07-19
lang-support: add list of packages that can be imported

aykevl:
Based on the work by @trashhalo:
https://github.com/trashhalo/tinygo-import-report


deadprogram:
This is very useful info! Now merging... thanks to @trashhalo for the original work, and @aykevl for the integration.

PR #33, opened 2019-07-31
Update bbc-microbit with info MacOS

abtris:
Update info for users on MacOS

### DIFF
--- a/content/microcontrollers/bbc-microbit.md
+++ b/content/microcontrollers/bbc-microbit.md
@@ -24,8 +24,12 @@ The micro:bit comes with the [DAPLink bootloader](https://tech.microbit.org/soft
 
 - Plug your micro:bit into your computer's USB port.
 - The micro:bit board will appear to your computer like a USB drive.
-- Determine the path to the board, on Linux it will be something like `/media/[USERNAME]/[NAME OF THE BOARD]`.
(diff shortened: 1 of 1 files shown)


content/microcontrollers/bbc-microbit.md
```diff
@@ -24,8 +24,12 @@ The micro:bit comes with the [DAPLink bootloader](https://tech.microbit.org/soft
 
 - Plug your micro:bit into your computer's USB port.
 - The micro:bit board will appear to your computer like a USB drive.
-- Determine the path to the board, on Linux it will be something like `/media/[USERNAME]/[NAME OF THE BOARD]`.
-- Build your TinyGo program to the board in .hex format using `tinygo build -o=/media/[USERNAME]/[NAME OF THE BOARD]/flash.hex -target=microbit [PATH TO YOUR PROGRAM]` command.
+- Determine the path to the board
+    - on Linux it will be something like `/media/[USERNAME]/[NAME OF THE BOARD]`.
+    - on Mac it is `/Volumes/[NAME OF THE BOARD]`.
+- Build your TinyGo program to the board in .hex format
```
justinclift: "to the board" -> "for the board"

deadprogram: In this case it is both building for the board (target) as well as TO the board (destination directory). So I let it go, as is.

PR #37, opened 2019-08-10
usage: update micro:bit description

zolaemil:
Bring description of micro:bit example in line with how the code actually works.


conejoninja:
Not sure if this PR should go against *master* or *dev* as it's fixing some documentation for existing devices. IMO it's ok to work on *master*, but don't want to mess with the workflow.

Merging into *dev* will add this commit : https://github.com/tinygo-org/tinygo-site/commit/864e93ee0a611f3e2c650aa7b2274f9585c841d2 For me both are ok, what is your opinion @deadprogram  ?

deadprogram:
As of right now, both `dev` and master are the same. I am been using `dev` only when we are preparing for a new release, and there is new functionality that is not yet released.

PR #38, opened 2019-08-12
boards: add installation instructions for Arduino Uno

aykevl:
It's a bit of duplication (especially the `avrdude` part, which is mentioned twice on this page) but it may help people new to TinyGo.
I wasn't sure whether to include these error messages, they take up quite a bit of space but may be helpful to quickly diagnose the problem.

See https://github.com/tinygo-org/tinygo/issues/504


justinclift:
> I wasn't sure whether to include these error messages ...

It's good to have them there, and they'll be found for search engine queries of the error text + "tinygo" (etc). :smile:

deadprogram:
Somehow missed this PR before. Now merging, thanks @aykevl

PR #39, opened 2019-08-28
Update doc to reflect lack of armhf deb file

asachs01:
Releases for 0.7.1 do not include any packages for armf. This PR addresses that.


deadprogram:
AT least until we actually have a build of this, seems like a good idea. Merging, thanks!

PR #44, opened 2019-09-18
lang-support: update list of importable packages

aykevl:
Also, limit the max number of lines to 15 for excessively long outputs.


deadprogram:
Merging.

Issue #45, opened 2019-09-22
Update installation/getting started docs


conejoninja:
Getting started docs, for example: https://tinygo.org/getting-started/linux/ , use *dep ensure* command which is no longer used (tinygo uses modules)

PR #46, opened 2019-09-25
docs: add info for Go modules based source install

deadprogram:
This PR adds info for Go modules based source install which is the current set of source requirements since the 0.8.0 release.

It corrects issue #45


justinclift:
Can we kill this branch?  It's been merged. :smile:

PR #47, opened 2019-09-25
WIP doc: auto-generate API documentation for the machine package

aykevl:
Work-in-progress documentation for the machine package. Sharing it now for others to see, in case there is a better way to do this.

The build tags etc. should be provided by the `tinygo` command. I have some code prepared that can expose these things.

### DIFF
--- a/doc-gen/main.go
+++ b/doc-gen/main.go
@@ -1,0 +1,269 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"golang.org/x/tools/go/packages"
+)
+
+// PackageDoc contains all the information necessary to render the documentation
+// of a single package.
+type PackageDoc struct {
+	Target    string
+	Types     map[string]*Type
+	Funcs     map[string]*ast.FuncDecl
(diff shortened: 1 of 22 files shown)


deadprogram:
What is the status with this branch?

aykevl:
Updated. I'm not sure why it didn't work before, but now it does. It is able to generate documentation for a whole list of microcontrollers.

This is not yet finished, the docs probably need to be moved around a bit. Also, it should somehow be integrated with the "microcontrollers" sections but I'm not quite sure how, maybe as a subpage of every supported board? Any ideas?

This PR depends on the `target` subcommand: https://github.com/tinygo-org/tinygo/pull/692

doc-gen/main.go
```diff
@@ -0,0 +1,269 @@
...
+	for _, target := range os.Args[1:] {
+		fmt.Println("Generating documentation for:", target)
+		path := filepath.Join("..", "content", "machine", target+".md")
+		writeTargetDoc(target, path)
+	}
+}
+
+func writeTargetDoc(target, path string) {
+	// Get the important information from the compiler.
+	cmd := exec.Command("tinygo", "target", target)
```
deadprogram: This needs be updated to `info`, right?

aykevl: Fixed.

aykevl:
The question now really is: where to put the documentation? It would be nice to include the docs on the tinygo.org website with the next release.

deadprogram:
No longer needed, so closing.

PR #51, opened 2019-10-18
Update some macOS related instructions

ff4f01:
These are some suggestions to some macOS-related documents, that I found it could be better if updated.


deadprogram:
Thank you for the PR. Now merging.

PR #58, opened 2019-11-26
Machine doc moved

deadprogram:
Basically #47 but moved under "Microcontrollers" and with links from each board to its machine package docs.


doc-gen/Makefile
```diff
@@ -0,0 +1,37 @@
...
+all:
+	go run . $(DEVICES)
+
+
+#all:
+#	go run . pca10040 > ../content/doc/pca10040/machine.md
+#	go run . hifive1b > ../content/doc/hifive1b/machine.md
+#	go run . bluepill > ../content/doc/bluepill/machine.md
+#	go run . gameboy-advance > ../content/doc/gameboy-advance/machine.md
+#
```
aykevl: Oops looks like this is some old code that should really be removed.

doc-gen/Makefile
```diff
@@ -0,0 +1,37 @@
...
+    metro-m4-airlift \
+    circuitplay-express \
+    stm32f4disco \
+    circuitplay-express \
+    gameboy-advance \
+    itsybitsy-m4 \
+    nucleo-f103rb \
+    hifive1b \
+    pinetime-devkit0 \
+    x9pro
```
aykevl: I see you reordered some of these devices, any reason for that?
The order doesn't really matter, but I kept it at the same order as the smoke tests in the Makefile, for easier updating (maybe?).

deadprogram: I was going to put it all into alpha order, but instead I just added the missing ones in semi-random order.

doc-gen/Makefile
deadprogram: Yeah, I was wondering. :scissors:

PR #64, opened 2020-01-03
replace i8 with i8* for two extra parameters of non-exported functions

mewmew:


aykevl:
Good catch! Thank you for this contribution. Merging.

PR #66, opened 2020-01-08
basic: clarify Circuit Playground Express

stapelberg:
On a just-bought board, the instructions didn’t work because boards are shipped in MakeCode mode, which requires just one RESET press.


deadprogram:
Thank you for this clarification @stapelberg now merging.

PR #67, opened 2020-01-15
Add Return Value to "func add()"

pfandzelter:
Will not compile otherwise.

### DIFF
--- a/content/webassembly/webassembly.md
+++ b/content/webassembly/webassembly.md
@@ -16,7 +16,7 @@ func main() {
 // This function is imported from JavaScript, as it doesn't define a body.
 // You should define a function named 'main.add' in the WebAssembly 'env'
 // module from JavaScript.
-func add(x, y int)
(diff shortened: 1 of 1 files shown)


aykevl:
This looks good to me. Thank you for the fix!

@deadprogram I changed the base to dev, but when I did that a new commit appeared. Apparently one commit has been merged into the master branch instead of the dev branch (while I merged other commits into the dev branch). I'm not sure what's best to do now? Rebasing this PR on top of dev would perhaps be the cleanest option.

deadprogram:
Agreed, that is probably easiest.

pfandzelter:
There is actually a bit more on this page that doesn't quite add up or would need extension. I will try to extend the documentation here a bit when I can

deadprogram:
For now I will just merge this PR, and we can figure out the rest from there in additional PRs.

Thank you @pfandzelter

PR #68, opened 2020-01-18
compiler-internals: add page on interrupts

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/782

Also reorder other pages a bit, in order from "relevant if you're doing low-level stuff as a user" to "if you want to dig deeper into the compiler".


deadprogram:
Most excellent work @aykevl thank you for the explanations, they will prove most helpful.

Now merging.

PR #69, opened 2020-01-22
Added info on  tinygo drivers compatible with PyBadge

gmeader:


deadprogram:
Hi @gmeader thank you for the contribution.

I changed the branch to the `dev` branch, so this info will get picked up as part of our next release.

Thanks again!

PR #70, opened 2020-02-05
Making GOPATH command work on newer systems

LaPingvino:
When installing Go out of the box, setting a GOPATH isn't necessary any more, but the internal one is still being used for go get, which is in $HOME/go. To access GOPATH also when it is not set in the shell you use, use go env GOPATH instead of $GOPATH.


aykevl:
Great, thank you!

PR #75, opened 2020-02-27
Update arduino-uno.md

Algogator:
Make flash command a little more clearer


deadprogram:
Hi @Algogator thank you for the additional documentation.

I moved the PR over to the `dev` branch so it gets picked up for our next release, which should be coming soon. Now merging.

Thanks again!

PR #76, opened 2020-02-27
Update basic.md

Algogator:
Adds location of the example scripts


aykevl:
I do not think downloading the TinyGo source code is necessary. You can simply run `tinygo flash -target=arduino examples/blinky1` with a release build because the examples/blinky1 code is included in the TinyGo root.

Issue #78, opened 2020-04-11
bossac installation for ubuntu needs more lines.


bus710:
On the way to install bossa for nano33 IoT, I had minor hiccups to install bossac in my ubuntu setup.

From bossa repo I found this: 
- https://github.com/shumatech/BOSSA/issues/81

It basically said that two more packages are required:
```sh
$ sudo apt install libreadline-dev libwxgtk3.0-* 
$ git clone https://github.com/shumatech/BOSSA.git
$ cd BOSSA
$ make
$ sudo cp bin/bossac /usr/local/bin
```

PR #79, opened 2020-04-11
Update arduino-nano33-iot for bossa install

bus710:
Regard to https://github.com/tinygo-org/tinygo-site/issues/78.

I just added couple line for bossac installation in debian/ubuntu based system since it has some dependencies of wxgtk and readline libraries. 

But I am not sure if the originally guide was intentionally written in that simpler way for readability.
If so, please let me know and dismiss.


deadprogram:
Hi @bus710 this was strictly an omission, not intentional. Thank you for the contribution, now merging.

PR #81, opened 2020-04-15
Update the Getting Started docs to use v0.13 instead of v0.12

suda:


deadprogram:
Thanks @suda now merging.

PR #82, opened 2020-04-22
update stdlib lang-support for go 1.14

elliots:
i only ran the ones with existing compile errors, but this might be better automated with something like this? i could get this to output markdown if its of use.

```bash

std=`go list std`
for pkg in ${std//\\n/
}
do
  echo
  echo " - $pkg"
  echo "package main
import (
	_ \"$pkg\"
)
func main() {}
" > tmp.go
  tinygo build -target wasm -o tmp.wasm tmp.go
done
```


elliots:
Just saw there is already a script :)

PR #83, opened 2020-05-03
getting-started: fix the docs for recent Go, LLVM and Debian versions

aykevl:
The biggest change is that `GO111MODULE` isn't necessary anymore with a go.mod file in the root directory.


deadprogram:
Looks good, thanks @aykevl now merging.

PR #84, opened 2020-05-07
getting-started: add docs for building TinyGo on Windows

aykevl:
This is probably missing something but it's better than nothing. Most importantly: it may require adding some paths to the `Path` environment variable and I'm not entirely sure whether MinGW-w64 works (but installing MinGW-w64 is easier than TDM-GCC which I did test). I think it should work because MinGW is installed on the hosted Azure CI agents.

I've also fixed some errors in the macos and linux guide.

EDIT: also note that this was my first time successfully building TinyGo on Windows. Previously I always did it on Azure CI.

### DIFF
--- a/content/getting-started/linux.md
+++ b/content/getting-started/linux.md
@@ -191,9 +191,10 @@ sudo apt-get install build-essential git cmake ninja
 ```
 
 The following command takes care of downloading and building LLVM. It places the
-source code in `llvm-build/` and the build output in `llvm/`. It only needs to
-be done once until the next LLVM release. Note that the `export` lines are
-optional, but using Clang during the build speeds up the build significantly.
+source code in `llvm-project/` and the build output in `llvm-build/`. It only
+needs to be done once until the next LLVM release. Note that the `export` lines
+are optional, but using Clang during the build speeds up the build
+significantly.
 
 ```shell
 export CC=clang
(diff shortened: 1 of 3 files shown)


sago35:
I have created a PR #85 for this pr.

sago35:
For better reproducibility, it might be better to prepare package.config as follows and fix the version. When I tried it yesterday, the version looked like this

```
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="cmake" version="3.17.2" />
  <package id="git" version="2.26.2" />
  <package id="golang" version="1.14.1" />
  <package id="make" version="4.3" />
  <package id="mingw" version="8.1.0" />
  <package id="ninja" version="1.10.0" />
  <package id="python2" version="2.7.18" />
</packages>
```

sago35:
Other than what I wrote above, I think it's good.

aykevl:
I do not think pinning specific versions is very helpful. The exact version should not matter a lot, and we do no specify package versions for other operating systems either. They will also quickly get out of date.

Specifically for Python: as mentioned in #85, Python 3 has worked for me (I agree the error messages is not very clear).

aykevl:
(I should update this PR to just install MinGW)

deadprogram:
> (I should update this PR to just install MinGW)

agreed.

aykevl:
I have updated this to specify MinGW instead.

deadprogram:
Merging.

PR #85, opened 2020-05-09
Change python to python2

sago35:
#84 

LLVM needs python2, not python3.

```
./llvm-project/llvm/CMakeLists.txt:658

if( ${PYTHON_VERSION_STRING} VERSION_LESS 2.7 )
  message(FATAL_ERROR "Python 2.7 or newer is required")
endif()
```

### DIFF
--- a/content/getting-started/windows.md
+++ b/content/getting-started/windows.md
@@ -117,11 +117,11 @@ You will need to have the following programs installed on your Windows system an
   * GNU Make
   * CMake
   * Ninja
-  * Python
+  * Python2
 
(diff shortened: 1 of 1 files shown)


aykevl:
I have used Python 3 and that worked just fine. In fact, Python 2.7 has been deprecated for a while.

Have you tried Python 3?

sago35:
I tried again.

I was able to install it with python3.
The reason for the problem was the environment variables PYTHON_HOME and PYTHON_PATH.
When I removed it, it worked fine with python3.

sago35:
@aykevl 
This PR turned out to be a mistake and I want to close it.
If it is not a problem for you if I close it, then I will close it.

aykevl:
Okay, closing. Thank you for investigating.

PR #86, opened 2020-05-16
typo fix why-a-new-compiler.md

TGeoF:
typo fix


deadprogram:
Hi @TGeoF thank you for the fix. I switched the branch to `dev` so that it gets properly merged, which I am also doing now.

Thanks again!

PR #87, opened 2020-05-17
WIP pwm

aykevl:
I've been working on some good documentation for PWM. It's definitely not finished, but I think this is already really neat. It's an interactive way of explaining PWM ([Bret Victor style](http://worrydream.com/#!/MediaForThinkingTheUnthinkable)), which may be easier to understand.

Screenshot from 2020-05-17 23-20-07

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -30,3 +30,7 @@ weight = 30
 name = "<i class='fab fa-slack'></i> Slack"
 url = "https://gophers.slack.com/messages/CDJD3SUP6/"
 weight = 40
+
+[markup.goldmark.renderer]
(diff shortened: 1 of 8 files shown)


aykevl:
Updated a bit with some code samples (and an incomplete interactive counter/channel example). Definitely not finished, as the code hasn't even been written to implement this PWM but I wanted to make sure that explaining how this works is not too difficult.

ysoldak:
I love this page, can we probably re-vive it and add to the site?
I'll try to cross-check the examples with the current codebase, rebase on dev's head and probably add a servo example.

PR #88, opened 2020-05-24
microbit: add note for flashing with a SoftDevice present

aykevl:
See: https://github.com/tinygo-org/tinygo/pull/1134


deadprogram:
Merging.

PR #89, opened 2020-05-27
Fix wasm documentation for `export` annotation

moznion:
`go:export` -> `export`


aykevl:
Thank you, looks like I missed those.

PR #90, opened 2020-06-26
Update `CLI Flashing on Windows`

sago35:
The following changes make the CLI Flash method on Windows the same as
on other operating systems (linux, macOS).

https://github.com/tinygo-org/tinygo/pull/1084
https://github.com/tinygo-org/tinygo/pull/1085

### DIFF
--- a/content/microcontrollers/arduino-nano33-iot.md
+++ b/content/microcontrollers/arduino-nano33-iot.md
@@ -101,8 +101,6 @@ Once you have installed the needed BOSSA command line utility, as in the previou
 ### CLI Flashing on Windows
 
 - Plug your Arduino Nano33 IoT board into your computer's USB port.
-- Double tap the "RESET" button on the board.
-- Wait until the Arduino Nano33 IoT board appears as a serial drive.
 - Build and flash your TinyGo code using the `tinygo flash` command. This command flashes the Arduino Nano33 IoT with the blinky1 example:
 
     ```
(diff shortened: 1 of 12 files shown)


sago35:
I considered and worked with the following files as the target of my modifications.

```
$ git grep "flash-1200-bps-reset" *.json
arduino-nano33.json:    "flash-1200-bps-reset": "true"
circuitplay-bluefruit.json:    "flash-1200-bps-reset": "true",
circuitplay-express.json:    "flash-1200-bps-reset": "true",
clue_alpha.json:    "flash-1200-bps-reset": "true",
feather-m0.json:    "flash-1200-bps-reset": "true",
feather-m4.json:    "flash-1200-bps-reset": "true",
itsybitsy-m0.json:    "flash-1200-bps-reset": "true",
itsybitsy-m4.json:    "flash-1200-bps-reset": "true",
metro-m4-airlift.json:    "flash-1200-bps-reset": "true",
pybadge.json:    "flash-1200-bps-reset": "true",
pygamer.json:    "flash-1200-bps-reset": "true",
pyportal.json:    "flash-1200-bps-reset": "true",
trinket-m0.json:    "flash-1200-bps-reset": "true",
wioterminal.json:    "flash-1200-bps-reset": "true",
xiao.json:    "flash-1200-bps-reset": "true",
```


The following three files are not yet documented.
I will be creating a PR to add them to the documentation.

* pygamer
* wioterminal
* xiao

aykevl (approved):
This seems reasonable now that Windows is better supported, so LGTM. @deadprogram any comments?

deadprogram:
This is great, thank you @sago35 now merging.

aykevl:
It looks like the instructions are now basically identical for all operating systems, so we might want to merge these into one. That's less clutter for users and less maintenance work.

deadprogram:
I think we can do that with some shared content that gets pulled into all three places. I will look into how Hugo can handle this.

PR #93, opened 2020-07-02
Thing to help get up and running quicker with the bluepill

audstanley:
I had some issues with getting the inexpensive blue pill that I have running on with the tiny-go flash command.  Here is how I was able to fix it, and since a lot of people buy blue pills from Amazon, this should help to fix a ton of errors they may get along the way.

 Here is what I was getting before the fix
```sh
STANLEY ⛰  >>> tinygo flash -target=bluepill -ocd-output main.go
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
adapter speed: 1000 kHz
adapter_nsrst_delay: 100
none separate
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : clock speed 950 kHz
Info : STLINK v2 JTAG v29 API v2 SWIM v7 VID 0x0483 PID 0x3748
Info : using stlink api v2
Info : Target voltage: 3.160026
Warn : UNEXPECTED idcode: 0x2ba01477
Error: expected 1 of 1: 0x1ba01477
in procedure 'program' 
in procedure 'init' called at file "embedded:startup.tcl", line 506
in procedure 'ocd_bouncer'
** OpenOCD init failed **
shutdown command invoked

```

and here is what I got after the fix.
```sh
STANLEY ⛰  >>> tinygo flash -target=bluepill -ocd-output main.go
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport
Info : The selected transport took over low-level target control. The results might differ compared to plain JT
adapter speed: 1000 kHz
adapter_nsrst_delay: 100
none separate
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : clock speed 950 kHz
Info : STLINK v2 JTAG v29 API v2 SWIM v7 VID 0x0483 PID 0x3748
Info : using stlink api v2
Info : Target voltage: 3.161583
Info : stm32f1x.cpu: hardware has 6 breakpoints, 4 watchpoints
target halted due to debug-request, current mode: Thread 
xPSR: 0x01000000 pc: 0x0800069c msp: 0x20000800
** Programming Started **
auto erase enabled
Info : device id = 0x20036410
Info : flash size = 64kbytes
target halted due to breakpoint, current mode: Thread 
xPSR: 0x61000000 pc: 0x2000003a msp: 0x20000800
wrote 4096 bytes from file /tmp/tinygo841772696/main.hex in 0.133260s (30.017 KiB/s)
** Programming Finished **
** Resetting Target **
shutdown command invoked
```

### DIFF
--- a/content/microcontrollers/bluepill.md
+++ b/content/microcontrollers/bluepill.md
@@ -29,3 +29,44 @@ Programs are loaded onto the Bluepill with a separate hardware programmer such a
 - Plug your STLink v2 programmer into your computer's USB port.
 - Plug your Bluepill into the STLink v2 programmer using the Bluepill `SWIO`, `SWCLK`, `3V3` and `GND` pins.
 - Build and flash your TinyGo program using `tinygo flash -target=bluepill`
+
+## OpenOCD Possible Modification
+
+If you run the flasher, and you are getting the error **Warn : UNEXPECTED idcode: 0x2ba01477** you can mitigate the problem by editing the stm32f1x.cfg file within openOCD
+```sh
+> tinygo flash -target=bluepill -ocd-output main.go
+Open On-Chip Debugger 0.10.0
+Licensed under GNU GPL v2
+For bug reports, read
+        http://openocd.org/doc/doxygen/bugs.html
+Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
+Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
+adapter speed: 1000 kHz
+adapter_nsrst_delay: 100
+none separate
+Info : Unable to match requested speed 1000 kHz, using 950 kHz
+Info : Unable to match requested speed 1000 kHz, using 950 kHz
+Info : clock speed 950 kHz
+Info : STLINK v2 JTAG v29 API v2 SWIM v7 VID 0x0483 PID 0x3748
+Info : using stlink api v2
+Info : Target voltage: 3.160026
+Warn : UNEXPECTED idcode: 0x2ba01477
+Error: expected 1 of 1: 0x1ba01477
+in procedure 'program' 
+in procedure 'init' called at file "embedded:startup.tcl", line 506
+in procedure 'ocd_bouncer'
+** OpenOCD init failed **
+shutdown command invoked
+
+```
+
+
+```sh
+> sudo vim /usr/share/openocd/scripts/target/stm32f1x.cfg
+```
+
+Edit the line **set _CPUTAPID 0x1ba01477** and change the value to **set _CPUTAPID 0x2ba01477**.  This is usually a known problem with the chip on blue pills, since there are a lot of different chips out there.  This way, you can still write to those microprocessors.  
+
+## Other possible problems
+
+If you get an error on the line  **Info : STLINK v2 JTAG v29 API v2 SWIM v7 VID 0x0483 PID 0x3748**, you likely have the SWDIO/SWCLK lines in reverse.  This is also a known problem on some of the cheaper blue pills.  In some cases, the outer case of the st-link v2 is not labeled properly (and the pcb inside the case is labeled in reverse).  You wont hurt anything by switching the wires, and see if that fixes the problem.  It is better to try that than pry apart the st-link v2 casing.


content/microcontrollers/bluepill.md
```diff
@@ -29,3 +29,44 @@ Programs are loaded onto the Bluepill with a separate hardware programmer such a
...
+shutdown command invoked
+
+```
+
+
+```sh
+> sudo vim /usr/share/openocd/scripts/target/stm32f1x.cfg
+```
+
+Edit the line **set _CPUTAPID 0x1ba01477** and change the value to **set _CPUTAPID 0x2ba01477**.  This is usually a known problem with the chip on blue pills, since there are a lot of different chips out there.  This way, you can still write to those microprocessors.
```
deadprogram, line 68: If we could override the `CPUTAPID` value passed to OpenOCD to set it to `0` then OpenOCD will ignore the ID when flashing. That would be a better solution for users, no doubt.

@aykevl what do you think?

audstanley: I'll definitely test it out if you guys make the argument change.

audstanley: ... I'll see if I can make the change in my fork, and recompile tinygo and test it out today.

audstanley: Do you have any examples of the openocd command that (i think) flashes from /tmp/tingoNNNNNNNNNN/main.hex?  

I can see that the /usr/share/openocd/scripts/target/stm32f1x.cfg checks the variable CPUTAPID first, though when I change my /etc/environment to reflect CPUTAPID="the correct id for my st-link" ... /bin/tclsh doesn't seem to take ANY environment variables from my system.  So unless there is a way to set the var before executing openocd, I'm not sure that there can be a solution unless one just edits the /usr/share/openocd/scripts/target/stm32f1x.cfg manually.

This seems counter intuitive that the writer of openocd would check the variable in tclsh, and one couldn't pass variables into openocd before running, so I might be missing something here.

audstanley:
```
STANLEY ⛰  >>> echo $CPUTAPID
0x2ba01477
STANLEY ⛰  >>> /bin/tclsh
% echo $CPUTAPID
can't read "CPUTAPID": no such variable
% set CPUTAPID 0x2ba01477
0x2ba01477
% echo $CPUTAPID
0x2ba01477
%
```

audstanley: ..and yeah if there is a way to just add a flag to tinygo, like -openocd-cputapid=0x2ba01477
that would be amazing.

audstanley: ~~How do you feel about setting **all** the _CPUTAPID's to zero?  Would that be a workable solution?  Might that break something?  I'm running on an inexpensive st-link v2, and would need to buy a jtag to see if it breaks it for all the people that use jtags.  I have no idea for sure, but I'm willing to give it a shot.  I'll order one on Amazon RN.~~

~~I could elevate user privileges (to root [at least on linux]) on a regex error match, and then the problem would only need to be fixed one time for all the boards.  Setting _CPUTAPID to zero worked perfectly btw.  Thanks for the suggestion.  I was thinking about global variables, but that is a way better solution.~~

~~I think a solution in tiny-go code would be an awesome opportunity for me.  Offering a solution in the docs is fine, but I would love to give the code base some tweaks. Let me know, and I'll run some tests with the more expensive st-link /w jtag and see if setting ALL the _CPUTAPID's breaks anything along the way. I think most people aren't going to buy the $30 st-link w/ jtag anyway... they're going to likely get the cheap one that's $2 for smaller projects.~~

I figured out how to do this logically without needing to elevate user privileges after diving into openocd even more.

aykevl: Strange, I have been able to flash the bluepill in the past without modifying any file. Although, the last time I tried it didn't work (don't remember why).

Yes, I believe it is possible to modify a variable on the command line. For some boards that use a J-Link there is a workaround, see:

https://github.com/tinygo-org/tinygo/blob/a9ba6ebad9ad85cacd9674a1af9229489a811f68/compileopts/config.go#L264-L270
```go
	if c.Target.OpenOCDTransport != "" && c.Target.OpenOCDTransport != "swd" {
		return nil, fmt.Errorf("unknown OpenOCD transport: %#v", c.Target.OpenOCDTransport)
	}
	// [...]
	if c.Target.OpenOCDTransport != "" {
		args = append(args, "-c", "transport select "+c.Target.OpenOCDTransport)
	}
```

This inserts a custom command (`transport select swd`). I think you can also set a variable that way. For example, we could add a `"cputapid"` key to the JSON file that if set, will modify that variable (possibly to 0).

audstanley: Right now, I've written into the main.go to regex match errors given from openocd, and rerun the openocd subprocess.  Once ever fifty flashes (or so, this is somewhat random)... openocd wont flash the device at all. I found that this has to do with how the st-link performs a software reset to the device, and I've provided a detection and fix for this also.  Replicating the issue is tough.  I have to keep re-flashing until I get the error, and test the fix.  I think I'm going to do a pull request soon.  Right now, I've written the code in main.go, and if there is another place/file you prefer the code changes, I am totally open for suggestions.

I would like to learn how to add support for the nucleo-f303, but I think that's going to have to be another pull request, until I learn more of the code base.

I'll make the pull request today, so you can see where I'm at.

audstanley: pull request is here : https://github.com/tinygo-org/tinygo/pull/1217

aykevl: This sounds like a bug in OpenOCD. And because it's not so frequent (and there is a workaround), I'm not sure whether we should be working around it in such a way. Generally trying to recognize errors from an error log with a regex seems way too brittle (it can break with an update, for example).

audstanley: The error of the CPUTAPID could be fixed by just prepending -c "set CPUTAPID 0" before the board cfg (awesome, easy)  Since tinygo makes a subprocess call to openocd, I would argue that tinygo is responsible for errors that occur while making these calls. I agree, regex is not the best solution, and a better solution would be if openocd had specific error codes that we could just cross check on an exit against some error code table (that might exist, I'll check, but I doubt it).

The other error from the software reset pin/button is not an openocd error. It's an error with the st-link being unable to reset connected device (via software).  As I was trying to make a simple fix for the CPUTAPID, I had gotten this new problem that kept me from just writing tinygo code several times.  I think that is unfortunate.  I've just been wanting to simply write code, and I've been slammed with trying to streamline.  The error is significant enough that it can stop someone from working with their blue pill for long periods with this error that becomes consistent, and maybe abandon the blupill all together with tinygo if it becomes frustrating enough.  If there is a regex performance hit from Tiny-Go it's only going to be when flashing with openocd, I think it's relatively small, and worth it under the circumstances.

I think there might be a better solution too.  I'm just not sure what it is.  I at least wanted to provide something that works for now, and if you're going to call openocd as a subprocess you're always going to run the risk of a break in the future, why not make it much easier for newcomers?

I like the idea of adding to the bluepill.json, and maybe changing the args for openocd. I just bought a brand new official $30 st-link and I can't even get the thing to flash at all right now... which is a bummer because I was hoping that might explain why the CPUTAPID has a specific address that openocd sets.  I'll dive a little deeper and see what I can find if I can even get the new st-link to work (I bought the damn thing from ST and I am having nothing but problems).

Another option might be to add an additional flag to Tiny-Go that will allow for higher level error handling?  -ocd-error-handling .  My goal is to flash raw stm33f103's, so having an error handler in the language would be nice if that is something that could be considered. 

Thank you guys for everything so far.

audstanley: @aykevl What device are you using to flash the bluepill? Are you using the "official" st-link, the more expensive j-link, or the cheaper st-link?

aykevl: I have used one of the cheap ST-Link v2 clones you can find on AliExpress. I believe they have the official firmware.

audstanley: @aykevl thanks again, closing here, and will submit another pull request once I've had a chance to edit the json and fiddle with the bluepill json configuration and openOCDConfig constructor.

PR #99, opened 2020-07-23
all: changeover to eliminate all direct use of master/slave terminology

deadprogram:
This PR is part of the changeover to eliminate all direct use of master/slave terminology.

Also see https://github.com/tinygo-org/tinygo/pull/1216


aykevl:
Looks good to me!

PR #102, opened 2020-08-16
Update docker image version in wasm instructions

rista404:


deadprogram:
Thanks for the correction @rista404 now merging.

PR #103, opened 2020-08-20
fix path of wasm page

conejoninja:
This fixes the "Webassembly page" link in https://tinygo.org/getting-started/using-docker/


deadprogram:

PR #104, opened 2020-08-22
added some media from 2020

conejoninja:


deadprogram:
@conejoninja I have been putting the posts in reverse order, so that the newest ones are on top. What do you think?

conejoninja:
I think I should put more attention, not sure what was in my mind because I checked the dates. I even thought that 2019 was unusual empty since the "last" date was February 2019 :rofl:

deadprogram:
Thanks, now merging!

PR #109, opened 2020-09-02
Run stdlib tests in parallel, for improved performance

aykevl:
On my somewhat old laptop (2 physical cores, 4 logical cores), this almost halves testing time. Systems with more cores should see a bigger drop.


deadprogram:
I forgot this was even here, sorry about that.

Merging.

aykevl:
Thank you!

PR #110, opened 2020-09-03
Standard library tests

aykevl:
~This is a draft because only a handful of packages actually pass the tests with local patches to TinyGo. Those patches should soon end up in a TinyGo PR. This PR is only ready when a reasonable number of packages pass tests: most fail right now not because the package doesn't work but rather because `tinygo test` is incomplete.~

### DIFF
--- a/content/docs/reference/lang-support/stdlib.md
+++ b/content/docs/reference/lang-support/stdlib.md
@@ -7,159 +7,161 @@ The following table shows all Go standard library packages and whether they can 
 
 Note that the fact they can be imported, does not mean that all functions and types in the program can be used. For example, sometimes using some functions or types of the package will still trigger compiler errors.
 
-Package | Importable
---- | --- |
-archive/tar |  <span style="color: green">✔</span> yes  | 
-archive/zip |  <span style="color: green">✔</span> yes  | 
-bufio |  <span style="color: green">✔</span> yes  | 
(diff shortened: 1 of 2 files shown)


dkegel-fastly:
This looks really good as is.  I can enhance it after it lands.

The only nit I can see is: when a package has no tests, it should probably get an n/a rather than a yes in the "passes tests" column.

aykevl:
Marked the PR as ready for review.

> The only nit I can see is: when a package has no tests, it should probably get an n/a rather than a yes in the "passes tests" column.

Agreed. Not entirely sure how to implement this though, currently it just looks whether the `tinygo test` process exited with an error or not. I guess this can be implemented later.

aykevl:
Another possible enhancement: run the tests for various platforms (Linux, `wasi`, `cortex-m-qemu`, maybe `wasm` or other OSes in the future). And display it in a nice table, of course.

dkegel-fastly:
I was thinking of just exposing the tests that are enforced by CI;
"make report-stdlib-tests-pass" in 
https://github.com/tinygo-org/tinygo/pull/2659
generates a nice table of those vs darwin, linux, wasi, and windows.

dkegel-fastly:
Oh, also, we might want to put an asterisk on the "passes tests" column to say that not all functions may be implemented yet... thinking of net and reflect, for instance.

aykevl:
This is now supported.

PR #112, opened 2020-09-07
ide-integration: update to use proper GOROOT

aykevl:
This depends on https://github.com/tinygo-org/tinygo/pull/1362.

### DIFF
--- a/content/ide-integration/_index.md
+++ b/content/ide-integration/_index.md
@@ -3,133 +3,51 @@ title: "IDE Integration"
 weight: 8
 ---
 
+IDEs need to have certain environment variables set before they work with TinyGo: `GOROOT` and `GOFLAGS`. You can determine the correct values from the `tinygo info` command (starting with TinyGo 0.15).
 
-## Downloading the source
-
-Either git clone the tinygo source from.
-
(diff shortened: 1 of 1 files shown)


sago35:
I tried this with vim.
This is a good and simple way for non-developers to check the source code.
I like that it's very easy to do.

However, a developer editing a machine package, for example, would need to know how to do the original.

aykevl:
The cached GOROOT will contain symbolic links, so any edit to the machine package is also an edit to the cached GOROOT. So, this also works for developers of TinyGo.

You only need to change the settings with uncommon updates, such as when you move the TinyGo location or when you update the Go version.

sago35:
Cool!

deadprogram:
Is this ready to move out of draft yet?

sago35:
This approach only works with recent dev branches.
I think it won't be possible to merge into a release branch until tinygo 0.15 is released.

sago35:
@deadprogram  @aykevl 
tinygo 0.15 has been released, so I guess it's now mergeable.

deadprogram:
OK, merging. Thanks!

PR #113, opened 2020-09-11
Updates TinyGo windows install guide to include Scoop instructions

thushan:
Updates the Windows installation guide to include a 'Quick Install' option to install via [Scoop](https://scoop.sh/) (Winget packages are also coming but this is the first step). Sticks to the same style as other documentation to be consistent.

Feedback welcome.

### DIFF
--- a/content/getting-started/windows.md
+++ b/content/getting-started/windows.md
@@ -11,31 +11,41 @@ This page has information on how to install and use TinyGo on Windows 10.
 
 We now have a native install for Windows 10.
 
-VERY IMPORTANT NOTE: You cannot yet create Windows binary programs using TinyGo, only MCU and WASM targets.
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @thushan thanks for working on all this, very cool! I have switched the branch to `dev` will be merging so it does out as part of the next release, which is actually tomorrow so the tinygo scoop package will have its first update.

Anyhow, thanks again.

thushan:
Oops, thanks. @deadprogram 

I'll maintain the scoop packages as you release things, but happy to hear your thoughts if adding TinyGo's own Scoop bucket might be a better alternative so you can synchronise releases on day-0 (scoop merges take ~10d). Just like the HomeBrew tap.

Happy to get it setup.

Issue #105, opened 2020-08-24
Add to the documentation on how to use openocd (atsamd21 / atsamd51)


sago35:
https://github.com/tinygo-org/tinygo/pull/1310

deadprogram:
We need this now, @sago35 :smile_cat:

sago35:
https://github.com/tinygo-org/tinygo/issues/1320

PR #117, opened 2020-09-18
boards: add cli-flashing-with-openocd

sago35:
#105 

I' ve created a document for feather-m4 as a draft.
If there is no problem, I will update the other boards.


deadprogram:
I think you have a good approach going here, @sago35 please continue.

sago35:
Once the series of pull-requests for openocd have been merged, they need to be updated with the latest information.

deadprogram:
@sago35 is this info still needing to be added to the new site structure/pages?

sago35:
I would like to add a document in docs/guides that says to flash via debugger.
It is very efficient because it allows flashing even when the microcontroller is crashed.

https://tinygo.org/docs/guides/

PR #118, opened 2020-09-27
boards/machine: unify procedures for each OS

sago35:
Separated from #117


aykevl:
Thank you! This looks reasonable. There should be no reason now to specify a different procedure for each OS (if there is still a difference, that's a bug and should be fixed in TinyGo).

PR #121, opened 2020-10-12
Update linux.md for ArchLinux

wingedrhino:
Updated links for ArchLinux, since it is now in the community repository

### DIFF
--- a/content/getting-started/linux.md
+++ b/content/getting-started/linux.md
@@ -108,7 +108,7 @@ This should allow you to compile and flash TinyGo programs on an Arduino or othe
 
 ### Arch Linux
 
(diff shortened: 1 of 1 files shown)


deadprogram:
@johanbrandhorst is this correct?

johanbrandhorst (changes requested):
There are actually two versions - thank you for adding a link to the source-built one, but I think we should also include a link to the binary version (which is the one we control). Could you please add back the old one, in addition to this new one?

deadprogram:
This PR is out of date with the current site structure, but is the info in it still needing to be added?

ysoldak:
I propose close this PR, @deadprogram 
The change [released already](https://github.com/tinygo-org/tinygo-site/blob/release/content/getting-started/install/linux.md#arch-linux).

PR #122, opened 2020-10-22
Fix adafruit clue target name and link

elliots:


aykevl (changes requested):
While this probably fixes things, this is not the right fix. See https://github.com/tinygo-org/tinygo/pull/1456 for how this should be fixed.

This might require a change to the Makefile (to replace `clue_alpha` with `clue-alpha`), feel free to make a PR for that or otherwise I can do that.

elliots:
Solved by https://github.com/tinygo-org/tinygo/commit/7a78b2dc0e0d051cfbcc9f9636e10a59cc7a7d67

PR #124, opened 2020-10-24
Add link to VSCode extension

elliots:
Extension worked great for me, though I don't know if there are other reasons it hasn't already been linked to.


deadprogram:
Thanks for the updated doc @elliots we had just forgotten to add it. Merging.

PR #125, opened 2020-10-27
microcontrollers: remove I2C from ESP8266

aykevl:
The ESP8266 does not have hardware support for I2C or PWM, this will need software emulation.

Many people may think the ESP8266 does have hardware support for I2C or PWM as it's available in all the other ports, but it does not. All of it is basically software bitbanging and using timers to simulate PWM behavior. But because it is in software, it is never as reliable as real hardware support.

The ESP32 has all these things, however.


sago35:
Please rebase to dev branch.

deadprogram:
This was done in #137 so now closing.

Issue #132, opened 2020-11-19
Add documentation for the board as added in 0.16.0


sago35:
* d1mini: add this ESP8266 based board
    * https://github.com/tinygo-org/tinygo/pull/1444
* esp32: use board definitions instead of chip names
    * https://github.com/tinygo-org/tinygo/pull/1439
* qtpy: add board definition for Adafruit QTPy
    * https://github.com/tinygo-org/tinygo/pull/1488
* teensy40: add this board
    * https://github.com/tinygo-org/tinygo/pull/1425
* wasi: add support for this target
    * https://github.com/tinygo-org/tinygo/pull/1373

aykevl:
All these are merged, so closing.

PR #135, opened 2020-11-19
boards: add QT Py to list of supported boards

sago35:
#132


deadprogram:
This was handled in #137 so now closing. Thanks @sago35

PR #136, opened 2020-11-22
Use Go modules

aykevl:
Instead of relying on GOPATH, use a Go modules instead. Hopefully that should fix some issues some people have.
But because that would interact badly with the go/packages package, disable Go modules while looking up those packages: otherwise github.com/tinygo-org/tinygo would be added as a dependency of this doc generation tool!


deadprogram:
I am still having the same problem with regenerating the pages. They each have the correct title, but the info itself is all removed.

aykevl:
For me it works: the contents is changed a bit but none of that is unusual. Without this PR I get the following error when compiling:

```
~/src/tinygo-site2/doc-gen$ make
go run . `tinygo targets`
../../golang.org/x/tools/internal/gocommand/vendor.go:16:2: cannot find package "golang.org/x/mod/semver" in any of:
	/usr/local/go/src/golang.org/x/mod/semver (from $GOROOT)
	/home/ayke/src/golang.org/x/mod/semver (from $GOPATH)
	/home/ayke/src/golang.org/x/mod/semver
make: *** [Makefile:2: all] Error 1
```

This is with Go version 1.15.5.

aykevl:
Oh I see where this is going wrong. I have added a commit with extra error checking, can you try it?

deadprogram:
Now it works as long as I have installed tinygo into the expected location. Thank you @aykevl now merging.

PR #138, opened 2020-11-27
Fix Adafruit Clue's target name

rsjethani:
Correct Adafruit Clue's target name to clue_alpha


aykevl:
We have standardized on dashes instead of underscores and are using that everywhere now. This should not be necessary in TinyGo 0.16.0 at least. See: https://github.com/tinygo-org/tinygo/pull/1456

Thank you for improving the website, but in this case we fixed the issue elsewhere for consistency.

PR #142, opened 2021-02-04
guides: add "guides" category for the machine package

aykevl:
This guide covers GPIO and I2C.

I plan on writing more. Eventually this should also cover PWM, for which I have already written some documentation (but not in this exact format).

Each guide has a certain structure:

 - short introduction
 - one or more sections in tutorial form
 - API documentation, in a form that is slightly more readable than the raw godoc (and of course split out by peripheral type to make it more focused)

I've written this for a few reasons:

- As an accessible introduction into various kinds of microcontroller peripherals. Godoc documentation is great, but the whole machine package can easily be overwhelming and you'll often need a bit more background to be able to use certain peripherals (just like you can't just do socket programming based simply on the API, you need some background on how sockets work).
- As cross chip documentation, explaining how chips are similar and different (example: not all chips have GPIO pull down mode).
- As a reference for implementing peripheral support for new microcontrollers, as a replacement for https://github.com/tinygo-org/tinygo/wiki/Machine-API which is wildly out of date.

I'm not entirely sure about calling them guides, maybe they should be called tutorials? Or something else? I'm open for suggestions.


aykevl:
Added I2C documentation.

deadprogram:
I think the idea of this is really great, overall.

Please see https://github.com/tinygo-org/tinygo-site/issues/144 to start a conversation about the subject of a possible site reorganization.

aykevl:
Yeah, I think what I wrote here is a mixture of how-to guide, reference and explanation content. I think it should have some more separation, I'll need to think about how exactly.

aykevl:
Superseded by #168.

PR #145, opened 2021-02-10
WIP: avr spi support

Nerzal:
Merge with next tinygo release, when AVR SPI support is being merged.

Just saw that i also need to update the machine package docs. Will do that later today.
https://tinygo.org/microcontrollers/machine/arduino/


deadprogram:
We will update machine docs when we do the release, as they are generated.

I switched this PR to `dev` branch, so it will be included in next release.

Thanks for all your work on this @Nerzal !

PR #146, opened 2021-02-13
add scheduler options

Nerzal:
Fixed some markdown lint warnings in the file. 
Also added the possible scheduler parameter. 
Some day in the future, someone who is able to explain the difference between the coroutines and task scheduler could add some more explanations there :)

### DIFF
--- a/content/usage/important-options.md
+++ b/content/usage/important-options.md
@@ -11,16 +11,16 @@ Output filename, see the ``build`` command.
 - `-target`
 Select the target you want to use. Leave it empty to compile for the host. This switch also configures the emulator, flash tool and debugger to use so you don't have to fiddle with those options. Read [supported targets](../../microcontrollers/) for a list of supported targets. Example targets:
 
- - wasm
+  - wasm
 WebAssembly target. Creates .wasm files that can be run in a web browser.
 
- - arduino
+  - arduino
 Compile using the experimental AVR backend to run Go programs on an Arduino Uno.
 
- - microbit
+  - microbit
 Compile programs for the `BBC micro:bit <https://microbit.org/>`_.
 
- - qemu
+  - qemu
 Run on a Stellaris LM3S as emulated by QEMU.
 
 - `-port`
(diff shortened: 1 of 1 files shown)


content/usage/important-options.md
```diff
@@ -67,3 +67,10 @@ Use the specified panic strategy. That is, what the compiled program should do w
 
   - `-panic=trap`
     Do not print the panic message but instead of printing anything, it directly hits a trap instruction. This instruction varies by platform but it will result in the immediate termination of the program. It could either exit with `SIGILL` or cause a call to the `HardFault_Handler`. It can be used to reduce the size of the compiled program while keeping standard Go safety rules intact at the cost of debuggability.
+
+- `-scheduler`
+Use the specified scheduler. Without a scheduler goroutines won't work.
+
+  - `scheduler=coroutines` Use the coroutines scheduler.
+  - `scheduler=tasks` Use the tasks scheduler.
+  - `scheduler=none` None is the default value. No scheduler is being used.
```
aykevl: > None is the default value

This is not true. The default scheduler varies by platform. For example, AVR currently defaults to `none` because it has such limited memory while `coroutines` and `tasks` is used for other platforms. Normally you do not need to override the default except on AVR where you can optionally select the tasks scheduler if you want concurrency.

The tasks scheduler is a scheduler much like an RTOS available for a limited number of platforms. It currently works on AVR, baremetal ARM (Cortex-M), the ESP8266, and the ESP32. This is usually the preferred scheduler if it is available.
The coroutines scheduler is a portable scheduler based on C++ coroutines that works almost everywhere and is therefore used whenever the tasks scheduler is not available or impossible to implement such as on WebAssembly.
The none scheduler disables scheduler support, which means that goroutines and channels are not available. It can be used to reduce firmware size and RAM consumption if goroutines and channels are not needed.

Nerzal: Thanks for the explanations, i'll update the pr :)

Nerzal: Well tbh your explanations are pretty good, can i just copy them?

aykevl: Sure :)

Nerzal: done :)

deadprogram:
Thanks for the docs updates @Nerzal now squashing/merging.

PR #149, opened 2021-03-02
Adds information about avr-gcc and avrdude.

thushan:
Adds information about where to find the `avr-gcc` and `avrdude` packages on Scoop to compliment the Arduino development.

Scoop packages are maintained by myself. Also addresses https://github.com/tinygo-org/tinygo/issues/1323.

### DIFF
--- a/content/getting-started/windows.md
+++ b/content/getting-started/windows.md
@@ -47,6 +47,23 @@ Upgrading to the latest TinyGo version can be done via scoop with:
 > scoop update tinygo
 ```
 
+#### Installing Arduino Dependencies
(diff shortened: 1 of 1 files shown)


aykevl:
Would it be possible to add the `avr-gcc` and `avrdude` packages as (optional) dependencies to scoop? That would avoid the need for documenting it making it easier for newcomers.

(Somewhere in the future we'll drop the avr-gcc dependency, but that probably still takes at least half a year)

deadprogram:
Thank you for the update. Now merging.

PR #150, opened 2021-03-04
getting-started: add `make wasi-libc` step

aykevl:
See: https://github.com/tinygo-org/tinygo/issues/1678


deadprogram:
Thanks for updating @aykevl now merging.

PR #154, opened 2021-03-10
Start with reorganization

aykevl:
This is a major reorganization of installation instructions.

  * "Build from source" instructions are moved into a separate page.
    This frees up space on the quick install pages. The reasoning is
    that these quick install pages will be used by more people and it's
    best to avoid as much clutter as possible.
  * A new Overview page is added. This gives a quick introduction into
    what TinyGo is and why this project exists, without going into
    details on why running Go on microcontrollers is a good idea (that's
    left for a future page).
  * I've removed Docker references from most install pages as it is
    already referenced in the mail install page
    (getting-started/install).
  * I've removed WSL from the Windows page because running TinyGo on
    Windows should be pretty solid by now and probably easier than
    running it in WSL.

Future goals:

  * Adding a blinky tutorial to the getting started pages.
  * Adding more guides, for example how to implement a device driver.

Apart from these changes, I've also moved the Linux and WebAssembly pages under the Guides category, which seems like a more appropriate place to me.


aykevl:
@deadprogram if you have time, it would be great if you could take a look at this PR.

aykevl:
Rebased after #159. @deadprogram can you maybe take a look?

deadprogram:
Been taking me a while, but I am really liking this. There will be more to do etc. but this gets us started.

Now merging.

PR #155, opened 2021-03-10
getting-started: add blinking LED tutorial

aykevl:
This is intended as the first tutorial anybody might want to follow when they're new to TinyGo. In fact, I've tried to make it so that non-Go programmers should also be able to follow along.

~~WIP because it depends on #154 (it needs a small update after that PR is merged).~~

### DIFF
--- a/content/docs/guides/_index.md
+++ b/content/docs/guides/_index.md
@@ -1,6 +1,6 @@
 ---
 title: "Guides"
-weight: 1
+weight: 2
 description: >
   These guides take you through the steps required to perform specific tasks using TinyGo.
 ---
--- a/content/docs/tutorials/_index.md
+++ b/content/docs/tutorials/_index.md
@@ -1,0 +1,6 @@
+---
+title: "Tutorials"
+weight: 1
+description: >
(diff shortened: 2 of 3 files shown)


aykevl:
Rebased on dev branch, it's now ready for review.

deadprogram:
Actually, I think we should add a new section `docs/tutorials` for this.

aykevl:
I was basing myself on the Django documentation, that puts installation instructions and tutorials in the same category: https://docs.djangoproject.com/en/3.2/intro/

Maybe it makes sense to split the two. Honestly it never made much sense to me that many projects put installation instructions next to tutorials, as if there is only one way to install software (especially with Linux package managers). So, not sure what to do here:

1. Split installation instructions (now "Getting started") and tutorials. Keep installation instructions separate from the rest of the documentation.
2. Move installation instructions under the wider "docs" umbrella, and let "getting started" be both installation instructions and tutorials of various kinds.

I'm fine with either option.

aykevl:
I have updated this PR with a new `docs/tutorials` section.

deadprogram:
:book: thanks @aykevl now merging.

PR #157, opened 2021-03-21
Flashing with openocd and nrf tools

oflebbe:
Adding more info how to flash nordic semi boards.

### DIFF
--- a/content/microcontrollers/nrf52840-mdk.md
+++ b/content/microcontrollers/nrf52840-mdk.md
@@ -23,12 +23,30 @@ The [nRF52840-MDK](https://wiki.makerdiary.com/nrf52840-mdk/) (not to be confuse
 
 ## Flashing
 
-### OpenOCD
+There are two options to flash the nRF52840-MDK board.
 
-Programs are loaded onto the nRF52840-MDK board using the `openocd` command line utility program. You must install [OpenOCD](http://openocd.org/) before you will be able to flash the nRF52840-MDK board with your TinyGo code.
+### OpenOCD (Recommended)
+
+You must install [OpenOCD](http://openocd.org/) before you will be able to flash the nRF52840-MDK board with your TinyGo code.
+You should check [OpenOCD Documentation](http://openocd.org/Documentation) for installation and configuration instructions.
+
+Once you have installed it correctly, you will be able to flash the nRF52840-MDK board with your TinyGo code.
 
 - Plug your nRF52840-MDK into your computer's USB port.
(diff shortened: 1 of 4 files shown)


content/microcontrollers/nrf52840-mdk.md
```diff
@@ -23,13 +23,31 @@ The [nRF52840-MDK](https://wiki.makerdiary.com/nrf52840-mdk/) (not to be confuse
...
+
+First install the J-Link Software and Documentation Pack from Segger: [https://www.segger.com/downloads/jlink/#J-LinkSoftwareAndDocumentationPack](https://www.segger.com/downloads/jlink/#J-LinkSoftwareAndDocumentationPack)
+
+Then install the nRF5x Command-Line Tools: [https://docs.zephyrproject.org/latest/guides/tools/nordic_segger.html#nrf5x-command-line-tools-installation](https://docs.zephyrproject.org/latest/guides/tools/nordic_segger.html#nrf5x-command-line-tools-installation)
+
+Once you have installed both of these correctly, you will be able to flash the nRF52840-MDK board with your TinyGo code.
+
+- Plug your nRF52840-MDK into your computer's USB port.
+- Build and flash your TinyGo program using `tinygo flash -target=nrf52840-mdk -programmer command`
+
```
sago35, line 50: This method probably won't work because flash-command is not set in nrf52840-mdk.json.

In my environment, feather-nrf52840 also gave an error because the flash-command is not set.
If I add the following to feather-nrf52840, I can write with --programmer command.

> "flash-command": "nrfjprog -f nrf52 --sectorerase --program {hex} --reset",

Perhaps it is better to set the flash-command in nrf52840.json and nrf51.json, and remove the setting from the json of each board.
We'll create a PR later.

If the PR is merged, then the document you have written is correct.

sago35 (approved):
The description in nrf52840-mdk.md is different from the current state, but I think the PR will be merged sooner or later, so there should be no problem.

content/microcontrollers/nrf52840-mdk.md
oflebbe: @sago35  Oh sorry, I didn't doublecheck that the nrf52840-mdk has the command in json file declared as well, since I only looked for the same text in documentation.
And btw, thanks for hinting me at WinUSB in openocd documentation, and the tools to set it. I wasn't aware of these dark corners in Win10.

deadprogram:
OK merging. Thanks everyone!

PR #158, opened 2021-03-26
Update Digispark's interfaces table

erazemk:
According to the Digispark documentation, the Digispark board supports PWM, so this commit updates the interfaces table to reflect that.


aykevl:
Indeed, it has PWM support. Thank you!

PR #163, opened 2021-04-23
Update WebAssembly documentation

surma:
The WebAssembly example page declares a function called `add` and mentions that it needs to get added to `env` in the imports object.

The instantiation code further down, however, doesn’t do so. This PR fixes that, but I am not entirely sure if this is the style you want to advocate for.

### DIFF
--- a/content/webassembly/webassembly.md
+++ b/content/webassembly/webassembly.md
@@ -84,6 +84,9 @@ loading it into JavaScript with `WebAssembly.instantiateStreaming`, or
 const go = new Go(); // Defined in wasm_exec.js
 const WASM_URL = 'wasm.wasm';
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @surma ffirst of all thanks fo rthe contribution.

We have just merged in a major revamp of the site docs in the `dev` branch which is where we are working on the next version of the site.

The file you changed in this PR is now located at https://github.com/tinygo-org/tinygo-site/blob/dev/content/docs/guides/webassembly.md

What do you want to do about this update, which still seems to me to be needed?

surma:
No, the new page shows how to populate the `importObj`, so we are all good on that branch!

Cheers :D

PR #164, opened 2021-05-04
Theme change and site reorganization

deadprogram:
This PR does 2 large things.

The first is to change the site to use the Docsy theme: https://www.docsy.dev/ 

The second furthers the reorganization started in #154 but then goes a lot further towards using the documentation system described in https://documentation.divio.com/

Here is what the current WIP looks like for the home page:

image

Here is what it looks like for the `getting started` section:

image

Here is what it looks like in the `docs` section:

image


aykevl:
Shiny!

I'll take a closer look soon.

aykevl:
This PR adds two files, assets/icons/logo.svg and static/images/tinygo-logo-10.svg, that aren't really SVG files. They are really just PNG files wrapped in a SVG container, so take up more space than a plain PNG files and provide none of the benefits of SVG (such as their scalability). I think these should be either real SVG files or instead the PNG versions should be used instead.

Other than that, this PR looks good to me.

aykevl:
The "IDE integration" menu item is slightly more bold than other pages at the same level. It appears this is because it is a directory, not a regular file (like content/docs/guides/linux.md). Unless you plan on adding more sub-pages, I think this should be a regular file (content/docs/guides/ide-integration.md).

deadprogram:
> This PR adds two files, assets/icons/logo.svg and static/images/tinygo-logo-10.svg, that aren't really SVG files. They are really just PNG files wrapped in a SVG container, so take up more space than a plain PNG files and provide none of the benefits of SVG (such as their scalability). I think these should be either real SVG files or instead the PNG versions should be used instead.
> 
> Other than that, this PR looks good to me.

This is due to how docsy handles logos: https://www.docsy.dev/docs/adding-content/iconsimages/#add-your-logo

I did not want to dig too deep right now to "fix" it, do I just did a "convert" and went with it. I would like a real SVG of the logo at some point.

deadprogram:
> The "IDE integration" menu item is slightly more bold than other pages at the same level. It appears this is because it is a directory, not a regular file (like content/docs/guides/linux.md). Unless you plan on adding more sub-pages, I think this should be a regular file (content/docs/guides/ide-integration.md).

I was actually thinking up breaking up into sub pages for each IDE, so make it easier to add screen shots etc.

aykevl:
> This is due to how docsy handles logos: https://www.docsy.dev/docs/adding-content/iconsimages/#add-your-logo

Well that's unfortunate. I would consider that a bug in Docsy. But not much we can do about it, then.

aykevl:
> I was actually thinking up breaking up into sub pages for each IDE, so make it easier to add screen shots etc.

Okay, makes sense!

aykevl:
Not an issue with this PR, but while we're at it: "Go on microcontrollers" is perhaps a bit of a misnomer. A better name might be "Supported microcontrollers" or even "Supported microcontroller boards".

deadprogram:
OK squashed into one giant commit, and ready to merge into `dev`.

We can release the new site at the same time as we push out the next release. In the meantime we can work on the `dev` branch using the new structure.

deadprogram:
Now merging!

PR #166, opened 2021-05-05
microcontrollers: rename boards to match their TinyGo target name

aykevl:
I think it's best to let the URL match the name that is used in the
TinyGo `-target` flag. Previously it was hard to change, but now that
the documentation is moved anyway we can fix this (URLs are 404 anyway).


deadprogram:
That seems like a very good idea. Auto-lookups/generation will be much easier this way.

Merging, thanks @aykevl

PR #168, opened 2021-05-05
Add GPIO and I2C documentation

aykevl:
This documentation comes from #142 but was updated to follow the new structure. The contents hasn't changed much, only the location.


deadprogram:
Great work, we needed info with this contents! Now merging.

PR #169, opened 2021-05-05
config: fix GitHub repo URL

aykevl:
This makes most links in the right sidebar work.


deadprogram:
Oops! That was error on my part. Thanks for the correction @aykevl now merging.

PR #170, opened 2021-05-06
microcontrollers: remove PWM support from the Mega 2560

aykevl:
The Arduino Mega 1280 has PWM support, but the Mega 2560 does not.


deadprogram:
:scissors: thanks. merging.

PR #171, opened 2021-05-06
Automatically update table of supported features per board

aykevl:
This should avoid outdated data in the table and avoids the need to manually maintain all of those.

Not all features are implemented yet (such as WiFi), but most are. More can be added in the future.


aykevl:
Updated to also include PWM.

deadprogram:
Is it correct that this PR should remove so many files such as `content/docs/reference/microcontrollers/machine/atmega1284p.md`?

aykevl:
Yes, that's correct. I have removed them in the first commit because that commit skips generating the machine package documentation for all the targets that don't have an accompanying board description (in content/docs/reference/microcontrollers/*.md). It would make sense to me to only keep the files that are actively updated.

deadprogram:
OK, was just checking. Merging, thanks for automating this process!

PR #172, opened 2021-05-08
content: correct broken links due to site reorganization

deadprogram:
This PR corrects a number of broken links due to site reorganization.


deadprogram:
Merging, we need to keep this process moving along. :factory:

PR #173, opened 2021-05-09
macOS bossa install doc fix

ysoldak:
`brew cask install` command was deprecated and now is completely [disabled](https://github.com/Homebrew/brew/pull/10056/files#diff-6b41cd34991fd4bf2887d97970060cd2c3303f29c13d51480dce512a9ebad369R177) and would not work

Shall use `brew install`


deadprogram:
Thank you very much for the update @ysoldak now merging.

PR #174, opened 2021-05-09
microcontrollers: redirect all original MCU pages to their new locations

deadprogram:
This PR uses the Netlify redirect capability to redirect all original MCU pages to their new locations.

See https://docs.netlify.com/routing/redirects/redirect-options/


deadprogram:
:arrow_down: :arrow_upper_left:

PR #179, opened 2021-05-13
concepts: explain how the package organization within TinyGo works

deadprogram:
Adds a page with diagram and explanations of how the packages are organized.

image


sago35 (approved):
I think it's very good.

deadprogram:
Thanks for the review @sago35 now merging.

PR #180, opened 2021-05-14
[DEV] Update list of supported packages

fgsch:
This is using tinygo 0.18.0 and go 1.16.4.

### DIFF
--- a/content/docs/reference/lang-support/stdlib.md
+++ b/content/docs/reference/lang-support/stdlib.md
@@ -9,8 +9,8 @@ Note that the fact they can be imported, does not mean that all functions and ty
 
 Package | Importable
 --- | --- |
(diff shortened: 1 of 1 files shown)


deadprogram:
I know it is not directly related to this PR, but the links to the what went wrong with the imports is broken.

The anchor `#crypto-ed25519` is now been rendered on the page as `#cryptoed25519`. Not sure if it is Docsy or newer versions of Hugo. Probably Docsy.

In any case, now that your fixes in #182 are merged and also my own change from 78d2ecde146b3a541fd7f827a168e7df0904ff6a this PR seems like no longer valid as we need to regenerate the files yet again.

Hence closing this PR in favor of a newly generated version. Thanks for helping get this all sorted out!

PR #182, opened 2021-05-14
Use Go modules

fgsch:
This would have a negative effect when looking for go packages so disable
it while we do so.  Also ensure we're running off from the correct path.


deadprogram:
Thank you for the corrections to the importer. Now merging.

PR #187, opened 2021-05-19
XIAO as a DAPLink debugger

ysoldak:
Some minimal info on DIY debugger option.


deadprogram:
Thank you for working on these docs. Now merging.

PR #189, opened 2021-05-30
Fix the number of menus displayed in the sidebar

### LINKED ISSUE #177: The menu on the microcontrollers page only shows 50 items
image

sago35:
Fixes #177 

image


deadprogram:
That list was so long, that somehow I missed it that entries were not there. Good catch, @sago35 and thanks for the fix!

Now merging into `dev` then I will deploy to `release`

PR #196, opened 2021-06-26
microcontrollers: add feather-nrf52840-sense

sago35:
see
https://github.com/tinygo-org/tinygo/pull/1969


deadprogram:
Thanks @sago35 now merging.

Issue #201, opened 2021-07-12
Build from source: Instructions could be clearer


soypat:
### Package key distro codename
Every time I try to install llvm I naively copy the below commands and end up messing up my local LLVM installation and eventually reinstalling ubuntu
```
echo 'deb http://apt.llvm.org/buster/ llvm-toolchain-buster-11 main' | sudo tee /etc/apt/sources.list.d/llvm.list
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
sudo apt-get update
sudo apt-get install clang-11 llvm-11-dev lld-11 libclang-11-dev
```

 think It'd be nice if we can replace it with something of the following sort:

_Add the package key, replacing with your distro's keyword, i.e buster for Debian, focal for Ubuntu 20.04_
```
echo 'deb http://apt.llvm.org/xxxxx/ llvm-toolchain-xxxxx-11 main' | sudo tee /etc/apt/sources.list.d/llvm.list
```
_Install packages_
```
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
sudo apt-get update
sudo apt-get install clang-11 llvm-11-dev lld-11 libclang-11-dev
```

### CGO_ENABLED
If you do not have CGO_ENABLED you'll get plenty errors which do not help narrow the underlying issue down.

This tip-off can be made explicit in instructions:

_After LLVM has been installed, installing TinyGo should be as easy as running the following command:_
```
CGO_ENABLED=1 go install
```
_CGO should be enabled or tinygo will fail to compile._

soypat:
Maybe add a table with data from https://apt.llvm.org/ to make it more explicit
i.e.
_Take care to replace `xxxxx` in the command with your distribution's codename_
| Distro | Version | Codename |
|---------|------------|----------------|
|Debian| 10 | `buster` |
|Debian| 11 | `bullseye` |
|Ubuntu| 20.04 | `focal` |
|Ubuntu| 18.04 | `bionic` |
|Ubuntu| 20.10 | `groovy` |
|Ubuntu| 21.04 | `hirsute` |
|Debian| sid | `unstable` |

Markdown attached
<details>

```
| Distro | Version | Codename |
|---------|------------|----------------|
|Debian| 10 | `buster` |
|Debian| 11 | `bullseye` |
|Ubuntu| 20.04 | `focal` |
|Ubuntu| 18.04 | `bionic` |
[...]

PR #202, opened 2021-07-12
clarify linux install from source instructions

soypat:
Solves #201


deadprogram:
Thanks for the improvements @soypat now merging.

aykevl:
Ideally people should actually read the instructions instead of blindly copying commands to their shell, but I agree this is the next best thing.

Issue #203, opened 2021-07-13
getting-started/linux: instructive adding user to dialout group/add permissions to /dev/ttyACM0


soypat:
Many new users to arduino if using tinygo will encounter this error https://askubuntu.com/questions/1056314/uploading-code-to-arduino-gives-me-the-error-avrdude-ser-open-cant-open-d. 

I propose we add a link or small instructive on how to add your user to permitted users for the device dialout. I am available to create the PR.

### From what I understand this is relevant info:

This error is usually encountered when flashing your first program, here's a typical error:
```
avrdude: ser_open(): can't open device "/dev/ttyACM0": Permission denied
```
The above error may be followed by `ioctl` errors too. 

To fix this we must make sure our user is part of dialout group and that the correct permissions are set on the port. 

First lets check the existence and permissions of the connected device (make sure your Arduino is connected):
```shell
 ls -l /dev/ttyACM0
```
In the author's case, this yields the output `crw-rw---- 1 root dialout 166, 0 jul 13 12:45 /dev/ttyACM0`. Notice the permissions are restrictive, we'd like all users to be able to read and write on this port.

To add read and write permissions on this port:

```shell
sudo chmod a+rw /dev/ttyACM0
```
The result of `ls` has now changed to `crw-rw-rw-+ 1 root dialout 166, 0 jul 13 13:08 /dev/ttyACM0`.

Let's also check if our user is in `dialout` group:
```shell
groups
```
Output:  `whoami adm cdrom sudo dip plugdev lpadmin lxd sambashare`

To add our user to dialout we write:
```shell
sudo usermod -aG dialout $(whoami)
```

aykevl:
If you want to add this to the FAQ, that would be appreciated.

However, regarding the solution: I don't think the `chmod` fix is ever a proper solution (it will probably stop working once you remove the device). Instead, adding yourself to the dialout group is the proper fix. We can perhaps make a list of Linux distributions and the group they use: perhaps other distributions use a different group name for /dev/ttyACM0 devices. So the solution would be: add yourself to the dialout group, log out and log in again.

soypat:
Solved by #204

PR #204, opened 2021-07-14
Create tinygo-flash-errors.md

soypat:
Please take a look. Solves #203


deadprogram:
Thanks for working on this @soypat I think this information would be better classified under "guides". Do you mind moving it over there?

soypat:
No problem, done!

soypat:
@deadprogram just squashed to minimize commit footprint

deadprogram:
Thanks for working on this @soypat now merging.

PR #207, opened 2021-07-29
search: add Algolia search to site

deadprogram:
This PR adds Algolia docs search to the site.

https://docsearch.algolia.com/

Looks nicer and works better than the Google Site Search which I also tried.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -26,6 +26,11 @@ unsafe = true
 [params]
 copyright = "The TinyGo Authors"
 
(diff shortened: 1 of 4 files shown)


sago35:
I think it's very good.

The search box will be hidden when the width of the top page is small.
It's difficult to balance this with the design. Is there a better way?

image
image
image

deadprogram:
You'd have to change Docsy to address that: https://github.com/google/docsy/

PR #208, opened 2021-08-08
tutorials: add PWM guide

soypat:
A guide on how PWM API looks like in TinyGo. We can agree that it may differ between boards. Maybe we could add `<details>` boxes with other board examples? 

<details><summary>Example of details box</summary>

```go
package main
```
</details>

Please take a look.


deadprogram:
Looks great! I changed the branch for the PR to `dev` so that it can go out along with the next release, which should be relatively soon.

Thanks @soypat now merging.

PR #213, opened 2021-09-11
Fix link to avr installation instructions

oflebbe:


deadprogram:
Thanks for the fix @oflebbe now merging.

PR #214, opened 2021-09-14
Update nano-rp2040 and nano33

ysoldak:
Plus link to lsm6dsox IMU documentation.


deadprogram:
Thanks for the updates @ysoldak now merging.

PR #217, opened 2021-10-20
machine: allow reading the pin state when it is configured as an input

aykevl:
See https://github.com/tinygo-org/tinygo/pull/2192


deadprogram:
:+1: merging.

PR #220, opened 2021-11-14
Fix broken links in build.md

girishramnani:


deadprogram:
Thanks for the corrections @girishramnani now merging.

PR #223, opened 2021-12-02
install/macos.md: mention how to install xtensa tools with brew

dkegel-fastly:
This helped me get past "make smoketest".


deadprogram:
Thanks @dkegel-fastly for the additional info. Now merging.

PR #225, opened 2021-12-15
reference (important-options): remove coroutines and extalloc

niaow:
This removes the coroutines scheduler and the extalloc gc, updating related information.

Don't merge until after https://github.com/tinygo-org/tinygo/pull/2381 and https://github.com/tinygo-org/tinygo/issues/2174


deadprogram:
Merging since both of these prereqs have been merged.

Issue #226, opened 2021-12-29
Need to add a document for 0.25.0


sago35:
I compared [tinygo-org/tinygo/README.md](https://github.com/tinygo-org/tinygo/blob/3e109fca5f289f5e97548f061d7e0e24c9727193/README.md) with tinygo.org.

I suggest the following two things to be done

1. Change to the notation in tinygo-org/tinygo/README.md
    * I think it will be less confusing for tinygo users.
    * It will be easier to check if you have enough documentation.
    * ex) change `Teensy 4.0` to `PJRC Teensy 4.0`
2. Add the missing documentation described below.

There is no documentation for the following targets, so it should be added.
For ESP32/ESP8266, it is probably better to modify tinygo-org/tinygo/README.md.

* [Arduino MKR WiFi 1010](https://store.arduino.cc/usa/mkr-wifi-1010)
* [Arduino Nano 33 BLE Sense](https://store.arduino.cc/nano-33-ble-sense)
* [BBC micro:bit v2](https://microbit.org/new-microbit/)
* [ESP32](https://www.espressif.com/en/products/socs/esp32)
* [ESP8266](https://www.espressif.com/en/products/socs/esp8266)
* [M5Stack](https://docs.m5stack.com/en/core/basic)
* [M5Stack Core2](https://shop.m5stack.com/products/m5stack-core2-esp32-iot-development-kit)
* [Raytac MDBT50Q-RX Dongle (with TinyUF2 bootloader)](https://www.adafruit.com/product/5199)
* [Seeed LoRa-E5 Development Kit](https://www.seeedstudio.com/LoRa-E5-Dev-Kit-p-4868.html)
* [ST Micro "Nucleo" WL55JC](https://www.st.com/en/evaluation-tools/nucleo-wl55jc.html)
* [The Things Industries Generic Node Sensor Edition](https://www.genericnode.com/docs/sensor-edition/)

sago35:
I will work on the following.

* Change to the notation in tinygo-org/tinygo/README.md 
* Add documents.
    * [Arduino MKR WiFi 1010](https://store.arduino.cc/usa/mkr-wifi-1010)
    * [M5Stack](https://docs.m5stack.com/en/core/basic)
    * [M5Stack Core2](https://shop.m5stack.com/products/m5stack-core2-esp32-iot-development-kit)

sago35:
The remains are as follows

[...]

PR #227, opened 2021-12-29
reference: unify notation for targets

sago35:
see #226

### DIFF
--- a/content/docs/reference/microcontrollers/hifive1b.md
+++ b/content/docs/reference/microcontrollers/hifive1b.md
@@ -1,5 +1,5 @@
 ---
-title: "HiFive1 RevB"
+title: "SiFive HiFive1"
 weight: 3
 ---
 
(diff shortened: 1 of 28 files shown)


sago35:
This is the same notation and order as tinygo-org/tinygo/README.md, which makes it easier to check for missing documentation.
This allows, for example, STM32 microcontrollers to be lined up close together.

image

content/docs/reference/microcontrollers/hifive1b.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "HiFive1 RevB"
+title: "SiFive HiFive1"
```
aykevl: The HiFive1 and HiFive1 Rev B are slightly different (incompatible) boards.

```suggestion
title: "SiFive HiFive1 Rev B"
```

aykevl (commented):
Looks reasonable to me, but see the comment below.

sago35:
Updated.

After this, create a PR that also modifies tinygo-org/tinygo/README.md to `SiFive HiFive1 Rev B` .

sago35:
@aykevl 
Please check it.

deadprogram:
Good cleanup here @sago35 thanks! Now merging.

Issue #228, opened 2021-12-29
doc-gen becomes panic


sago35:
It appears to be unable to process the rp2040 source.

```
$ git rev-parse HEAD
ae52920d36f1c973db853d9f545ab0ad2955f9c5

$ go run . pico
Generating documentation for: pico
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x10 pc=0x80a84c]

goroutine 1 [running]:
main.writeMachinePackageDoc({0xc00000a380, 0x3a}, {0xc0000140d0, 0x4}, 0xc000266c60)
        C:/Users/sago3/dev/src/tinygo.org/x/tinygo-site/doc-gen/main.go:232 +0x74c
main.updateDocumentation({0xc0000140d0, 0x4}, {0xc00000a380, 0x3a}, {0xc00000a3c0, 0x32})
        C:/Users/sago3/dev/src/tinygo.org/x/tinygo-site/doc-gen/main.go:148 +0x68a
main.main()
        C:/Users/sago3/dev/src/tinygo.org/x/tinygo-site/doc-gen/main.go:92 +0x3b8
exit status 2
```

sago35:
The problem with this issue has been resolved.

PR #229, opened 2021-12-30
doc-gen: skip unexported receiver

sago35:
This PR will skip the case where the receiver is private.

#228 had a panic in the following source code.
https://github.com/tinygo-org/tinygo/blob/dev/src/machine/machine_rp2040_i2c.go#L440-L444


aykevl (approved):
This looks correct, thank you!

PR #241, opened 2022-01-25
content: update version number everywhere for 0.22 release

deadprogram:
This PR updates version number everywhere for the 0.22 release


sago35 (approved):
@deadprogram 
Looks good to me.
I guess we need to re-run doc-gen.

deadprogram:
> I guess we need to re-run doc-gen.

Correct. I will look at that...

PR #243, opened 2022-01-28
Document adafruit products associated with LED controllers.

Elara6331:
Some of Adafruit's boards have LEDs they call DotStar or Neopixel. In TinyGo, the driver used for them uses the technical name rather than the one Adafruit uses, and it is not easily clear which driver should be used.

This PR adds notes to the microcontrollers page of each board with a Neopixel or DotStar LED.

tinygo-org/tinygo#2591


deadprogram:
Talk about atomic commits! Thanks @Arsen6331 for such a disciplined contribution. Now mergnig.

PR #244, opened 2022-01-29
Tips, Tricks and Gotchas

ysoldak:
Finally assembled an embryo of a long planned Tips&Tricks page.
We may want to call it "Best practices", though I'm not sure about that.

Please, suggest improvements or add other tips.

### DIFF
--- a/content/docs/guides/tips-n-tricks.md
+++ b/content/docs/guides/tips-n-tricks.md
@@ -1,0 +1,69 @@
+---
+title: "Tips, Tricks and Gotchas"
+weight: 4
+description: |
+  Tips and tricks for small places.
+  How to write efficient embedded code and avoid common mistakes.
+---
+
+## Ensure Concurrency
+
+TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
+Since scheduling in TinyGo is [cooperative](https://en.wikipedia.org/wiki/Cooperative_multitasking), a goroutine that never does IO or other blocking calls (f.ex. `time.Sleep()`) will lock the single available thread only for itself and never allow other goroutines to execute. In such cases, you can use `runtime.Gosched()` as a workaround.
+
+```
+package main
+
+import (
+    "fmt"
+    "time"
+    "sync/atomic"
+    "runtime"
+)
+
+func main() {
+
+    var ops uint64 = 0
+    for i := 0; i < 50; i++ {
+        go func() {
+            for {
+                atomic.AddUint64(&ops, 1)
+                runtime.Gosched()
+            }
+        }()
+    }
+
+    time.Sleep(time.Second)
+
+    opsFinal := atomic.LoadUint64(&ops)
+    fmt.Println("ops:", opsFinal)
+}
+```
+
+## Save Memory
+
+Use the fact [slices are descriptors of array segments](https://go.dev/blog/slices-intro#slice-internals) to save memory and avoid unnecessary allocations.
+Allocate once an array of maximum size you ever need and slice it in your methods to the required sizes as you go.
+
+```
+type MyType struct {
+    // ...other fields
+    buf [6]byte
+}
+
+func (t *MyType) SomeMethod() {
+    tmpBuf := t.buf[:2] // this method needs a buffer of size 2
+    // ... use the buffer
+}
+```
+
+You may even find yourself having two or more slices pointing at diferent regions of the same array simultaneously, if you are careful.
+
+```
+buf1 := t.buf[:2]
+buf2 := t.buf[2:6]
+```
+
+For more recipes on how to avoid or minimize allocations while working with slices, please see [additional tricks](https://github.com/golang/go/wiki/SliceTricks#additional-tricks) section of [slice tricks](https://github.com/golang/go/wiki/SliceTricks) page.
+
+Familiarize yourself with [heap allocation concept](../concepts/compiler-internals/heap-allocation.md) and use it to your advantage.


content/docs/guides/tips-n-tricks.md
```diff
@@ -0,0 +1,70 @@
...
+## Ensure Concurrency
+
+TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
+This means we almost always have less threads than goroutines.
+Since scheduling in TinyGo is [cooperative](https://en.wikipedia.org/wiki/Cooperative_multitasking), a goroutine that never does IO or other blocking calls (f.ex. `time.Sleep()`) will lock the single available thread only for itself and never allow other goroutines to execute. In such cases `runtime.Gosched()` shall be used to force scheduler update.
+
+```
+package main
+
+import "fmt"
```
soypat: These import statements should be grouped into one `import ( ... )` clause.

content/docs/guides/tips-n-tricks.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+## Ensure Concurrency
+
+TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
+This means we almost always have less threads than goroutines.
+Since scheduling in TinyGo is [cooperative](https://en.wikipedia.org/wiki/Cooperative_multitasking), a goroutine that never does IO or other blocking calls (f.ex. `time.Sleep()`) will lock the single available thread only for itself and never allow other goroutines to execute. In such cases `runtime.Gosched()` shall be used to force scheduler update.
+
+```
+package main
```
soypat, line 15: Maybe add a short description of what happens when we run this program on linux vs microcontroller?

soypat:
@ysoldak @deadprogram Do we want this? I think it's pretty cool and am thinking I might add more items to the list in the future.

ysoldak:
I'd say let's merge and improve as follow ups?

content/docs/guides/tips-n-tricks.md
aykevl: I don't think there is a lot of difference? They are both single threaded and use the same scheduler.

content/docs/guides/tips-n-tricks.md
```diff
@@ -0,0 +1,70 @@
...
+        }()
+    }
+
+    time.Sleep(time.Second)
+
+    opsFinal := atomic.LoadUint64(&ops)
+    fmt.Println("ops:", opsFinal)
+}
+```
+TODO adapt example to embedded case
```
aykevl: TODO is for writers, not for readers. I don't think we should be showing this TODO to users. If you really want to keep it, consider `<!-- TODO ... -->` (not sure whether this syntax works, but I would assume so)

That said, I expect this code to work as-is on most chips.

content/docs/guides/tips-n-tricks.md
```diff
@@ -0,0 +1,70 @@
...
+description: |
+  Tips and tricks for small places.
+  How to write efficient embedded code and avoid common mistakes.
+---
+
+## Ensure Concurrency
+
+TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
+This means we almost always have less threads than goroutines.
+Since scheduling in TinyGo is [cooperative](https://en.wikipedia.org/wiki/Cooperative_multitasking), a goroutine that never does IO or other blocking calls (f.ex. `time.Sleep()`) will lock the single available thread only for itself and never allow other goroutines to execute. In such cases `runtime.Gosched()` shall be used to force scheduler update.
```
aykevl: I'm not entirely sure this is good advice here. While this is technically correct, it may be better to first investigate whether it needs to lock up the CPU in the first place (using a mutex, channel, or something like that).
Something like this maybe?

> In such cases, you can use `runtime.Gosched()` as a workaround.

content/docs/guides/tips-n-tricks.md
aykevl: I don't think there is a lot of difference? They are both single threaded and use the same scheduler.

ysoldak:
I'll address existing comments later today.
But for new ones, just suggest changes, people? The GitHub way, easy to apply. Instead of commenting back and forth.

ysoldak:
Addressed comments and squashed. PTAL.

ysoldak:
Sorry for spam, probably, GitHub UI kills me, just wanted to re-request review from you guys, @aykevl and @soypat

soypat (approved):
LGTM! Love the slice tip

content/docs/guides/tips-n-tricks.md
```diff
@@ -0,0 +1,70 @@
...
+weight: 4
+description: |
+  Tips and tricks for small places.
+  How to write efficient embedded code and avoid common mistakes.
+---
+
+## Ensure Concurrency
+
+TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
+This means we almost always have less threads than goroutines.
```
aykevl: > This means we almost always have less threads than goroutines.

True (there could in theory be zero goroutines and one thread), but not sure it is helpful? I think removing this line might avoid some confusion.

aykevl (approved):
Looks good to me.

content/docs/guides/tips-n-tricks.md
ysoldak: I believe, I was citing your slack message. But yes, less confusing text, the better. I remove.

ysoldak:
@aykevl PTAL, removed confusing sentence.

deadprogram:
Overdue for merge. Thanks for being patient @ysoldak

PR #247, opened 2022-02-04
typo: bring missing "e" back

antonfisher:


deadprogram:
Thanks for the correction @antonfisher now merging.

PR #253, opened 2022-03-10
Update build from source instructions to llvm 13

ZauberNerd:
Tinygo bumped the default llvm version to v13 in:
https://github.com/tinygo-org/tinygo/commit/3a4e0c94e1cbad7227d311e652a0903b26e47273

Also, cgo is enabled by default, no need to set the env variable.


deadprogram:
Thank you for the updates @ZauberNerd now merging.

PR #258, opened 2022-04-21
compiler-internals: update inline assembly page after compiler updates

aykevl:
For details, see:
  * https://github.com/tinygo-org/drivers/pull/401
  * https://github.com/tinygo-org/tinygo/pull/2790

Essentially, this updates the inline assembly to the current best practice.


deadprogram:
Thanks @aykevl now merging

PR #262, opened 2022-05-01
linux.md: correct URL of ARM deb, add tip on running on 64 bit ARM os.

dkegel-fastly:
Improves doc for raspberry pi users, see https://github.com/tinygo-org/tinygo/issues/2815


deadprogram:
Thanks for the correction @dkegel-fastly now merging.

PR #272, opened 2022-06-29
Auto-generate a "Pins" section

aykevl:
This can simply be updated using `make` in the doc-gen directory.

It needs https://github.com/tinygo-org/tinygo/pull/2940 in TinyGo for doc-gen to work correctly.


aykevl:
This is now ready!

The pins are based on https://github.com/tinygo-org/tinygo/commit/0a93347e1c1611f5c5ebcb138759fd4d0e522c7d and not on the release branch, but I don't think that matters much: there are only a few small changes between them. It will become consistent again with the next release.

deadprogram:
:sewing_needle:

PR #277, opened 2022-07-12
Minimum go version is 1.16

fgsch:
Spotted by @Integralist.

### DIFF
--- a/content/getting-started/install/macos.md
+++ b/content/getting-started/install/macos.md
@@ -9,7 +9,7 @@ description: >
 
 This page has information on how to install and use TinyGo on macOS. If you wish to build TinyGo from source, for example if you intend to contribute to the project, please take a look [here](../../../docs/guides/build).
 
-You must have Go v1.15+ already installed on your machine in order to install TinyGo. We recommend Go v1.17+.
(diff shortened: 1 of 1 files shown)


content/getting-started/install/macos.md
```diff
@@ -9,7 +9,7 @@ description: >
 
 This page has information on how to install and use TinyGo on macOS. If you wish to build TinyGo from source, for example if you intend to contribute to the project, please take a look [here](../../../docs/guides/build).
 
-You must have Go v1.15+ already installed on your machine in order to install TinyGo. We recommend Go v1.17+.
+You must have Go v1.16+ already installed on your machine in order to install TinyGo. We recommend Go v1.17+.
```
deadprogram, line 12: We probably should also "recommend" `v1.18+`

fgsch: I can change it but the other versions are recommending 1.17+.

deadprogram: Maybe we should change them all? Probably yes.

fgsch: Sure. I will do that in a separate PR though.

deadprogram:
Thanks @fgsch

PR #282, opened 2022-08-07
Run `make llvm-source` before `make llvm-build`

abrender:
When following steps in the current documentation, the build will fail with:
```
$ make llvm-build
....
CMake Error: The source directory "$DIR/llvm-project/llvm" does not exist.
```

This is because `make llvm-source` must be run first, to download the sources.


aykevl (approved):

deadprogram:
Thanks @abrender now merging.

PR #284, opened 2022-08-20
Replace AUR package with Arch community repo package

Elara6331:
Arch has added TinyGo to the community repo. This PR replaces the link to the AUR package with a link to the package in the community repo. This is better for users, as not all users have an AUR helper installed, and some are running Arch-based distros that discourage AUR use.


deadprogram:
Thanks for pointing this out @Arsen6331 now merging.

PR #285, opened 2022-09-02
Fix broken link to the BBC micro:bit schematic pdf

justinclift:
Looks like they've moved the pdf into a subdirectory, so the link is updated to match.


sago35 (approved):
LGTM!

deadprogram:
Thanks @justinclift now merging.

PR #287, opened 2022-09-04
guides: add guide how to optimize binary size

aykevl:
This is mostly intended for WebAssembly but should be useful for all targets.

This has been requested a few times, so it's useful to have a central place where these flags are documented.


deadprogram:
This is great @aykevl thank you! Now merging.

PR #288, opened 2022-09-05
microcontrollers: list PWM peripherals along with the pins

aykevl:
Add PWM peripherals in the pin table so that it becomes easier to know which PWM peripheral is usable for a given pin.

This doesn't include all chips. Specifically:

  - stm32 chips aren't included yet
  - nrf chips can use any PWM peripheral for any pin, so a table doesn't make much sense

Future ideas:

  * Also list other peripherals, like SPI, I2C, UART, etc.
  * Maybe also make a table of supported peripherals? Not sure whether that's useful or just clutters the page more.

This PR requires https://github.com/tinygo-org/tinygo/pull/3134.


deadprogram:
This is very helpful! Thanks @aykevl

PR #289, opened 2022-09-08
concepts: add page with some porting hints

aykevl:
See for example https://github.com/tinygo-org/tinygo/issues/1284#issuecomment-1233151620

### DIFF
--- a/content/docs/guides/compatibility.md
+++ b/content/docs/guides/compatibility.md
@@ -1,0 +1,66 @@
+---
+title: "Porting code to TinyGo"
+type: "docs"
+description: >
+  How to port existing code to TinyGo.
+---
+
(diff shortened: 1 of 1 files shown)


deadprogram:
This is very helpful info. Just wondering if it should be under "Guides"?

aykevl:
Yeah, that's probably a better place. I put it there because it seems similar to the FAQ.
No strong opinion from me.

deadprogram:
A page that is a "how to" is likely a guide.

A page that is a "what is" is probably a concept.

At least that is what makes most sense to me.

aykevl:
Okay, I've moved the page.

deadprogram:
Great start, thank you @aykevl now merging.

PR #292, opened 2022-09-15
concepts: update goroutine information on datatypes page

aykevl:
The information there is way out of date. This gets it up to date with the current state of the compiler and runtime.


deadprogram:
Thanks for the updated info!

PR #293, opened 2022-09-17
guides: add how to build llvm on windows

sago35:
ccache is installed by choco install mingw.
The addition of CCACHE=OFF is necessary because if ccache is present in windows, it is currently not possible to build.


deadprogram:
Good catch on that @sago35 merging.

PR #294, opened 2022-09-17
microcontrollers: add rp2040 boards

sago35:
badger2040
challenger-rp2040
macropad-rp2040
qtpy-rp2040
trinkey-qt2040
tufty2040
xiao-rp2040


deadprogram:
That's a lot of boards that we added! :smile_cat: thanks @sago35

PR #295, opened 2022-09-17
microcontrollers: change to documentation with usbcdc assumption

sago35:
Now it is no longer necessary to press BOOTSEL.
Also, the default machine.Serial is now USBCDC.


deadprogram:
So many improvements! :heart:

PR #298, opened 2022-09-28
machine: add PinOutputOpenDrain and PinDisable

aykevl:
These two are used on the Teensy boards (teensy36, teensy40, teensy41) and seem quite sensible to me, but weren't documented before. It is my goal to let the public API of the machine package match the documentation here, so that's why I'd like to add these two pin modes.

They aren't supported on other chips/boards yet though, but that can change of course.

@ardnew @firelizzard18


firelizzard18:
LGTM

PR #303, opened 2022-10-21
Update build.md

soypat:
Missing the `-tags=llvm14` has caused me many headaches, but now it can be solved!


deadprogram:
Thank you @soypat now merging.

aykevl:
Yes, there will be a build error when you're on the dev branch (with LLVM 15 currently) but you have LLVM 14 installed. Installing LLVM 15 fixes this. That can be easy or hard depending on your OS (Debian/Ubuntu/MacOS are easy, other distros may be more difficult).
We should update the documentation for the next release.

PR #307, opened 2022-11-03
getting-started: remove xtensa-esp32-elf requirement

aykevl:
These aren't necessary anymore with LLD support for Xtensa.

Should be merged only after https://github.com/tinygo-org/tinygo/pull/3264 is merged.

### DIFF
--- a/content/getting-started/install/macos.md
+++ b/content/getting-started/install/macos.md
@@ -53,12 +53,3 @@ brew tap osx-cross/avr
 brew install avr-gcc
 brew install avrdude
 ```
-
-#### Xtensa ESP32
-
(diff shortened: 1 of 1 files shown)


sago35:
@aykevl 
There is information on the Espressif toolchain on the ESP32 and other pages, but is this no longer necessary?
Does this mean that only esptools.py is needed now?

https://github.com/tinygo-org/tinygo-site/blob/release/content/docs/reference/microcontrollers/esp32-coreboard-v2.md#cli-flashing-on-linux

deadprogram:
That is correct @sago35

deadprogram:
Thanks @aykevl and @sago35 now merging.

aykevl:
@sago35 you are correct! Here is a PR: https://github.com/tinygo-org/tinygo-site/pull/311

PR #311, opened 2022-11-17
microcontrollers: remove esp32 toolchain installation instructions

aykevl:
These are not necessary anymore.


deadprogram:
Thanks @aykevl now merging.

PR #312, opened 2022-11-22
fix broken link for lora sx1261

inindev:


deadprogram:
Hello, @inindev do you think you could please submit this PR against the TinyGo drivers repo? That is more or less the "source of truth" on those links, e.g. I usually copy/paste from there to update.

https://github.com/tinygo-org/drivers/blob/dev/README.md#currently-supported-devices

PR #317, opened 2022-12-21
reference/machine: added eeprom reference and machine docs

SoleimanyBen:
relates to https://github.com/tinygo-org/tinygo/pull/3306

Added machine API, reference, and peripheral concept docs for the `EEPROM`.

I used the other docs as examples on where I should be placing things, so if something is not in the correct location, just let me know!

The peripheral concept page that I wrote feels a little lackluster, so please let me know if I should change it or add something to it.

### DIFF
--- a/content/docs/concepts/peripherals/eeprom.md
+++ b/content/docs/concepts/peripherals/eeprom.md
@@ -1,0 +1,55 @@
+---
+title: "EEPROM"
+weight: 3
+description: |
+  How to use the internal EEPROM.
+---
+
+For API documentation, see the [machine API](../../../reference/machine#eeprom).
+
+The EEPROM is a form of non-volatile memory that is uusually integrated along side an MCU. EEPROMs allows you to store a relatively small amount of data by allowing bytes to be erased and reprogrammed. For the most part, EEPROMs are not integrated directly onto a board, but in some cases they are provided as a standard peripheral.
+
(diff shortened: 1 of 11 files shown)


content/docs/concepts/peripherals/eeprom.md
```diff
@@ -0,0 +1,55 @@
...
+weight: 3
+description: |
+  How to use the internal EEPROM.
+---
+
+For API documentation, see the [machine API](../../../reference/machine#eeprom).
+
+The EEPROM is a form of non-volatile memory that is uusually integrated along side an MCU. EEPROMs allows you to store a relatively small amount of data by allowing bytes to be erased and reprogrammed. For the most part, EEPROMs are not integrated directly onto a board, but in some cases they are provided as a standard peripheral.
+
+Although it may seem useful to have some type of data storage for your applications, the EEPROM is often limited in the amount of write cycles that it contains. This means that after a certain amount of writes, the EEPROM will not writeable, but you may still read data from it.
```
aykevl, line 12: EEPROMs are almost always included directly on the chip, I think?
Also, here is a suggestion for the text:

```suggestion
An EEPROM is a form of non-volatile memory that is usually integrated in an MCU. EEPROMs allow you to store relatively small amounts of data by allowing bytes to be erased and reprogrammed.

EEPROMs, like flash, can only be written a limited number of times but are not limited in how often they can be read. Therefore, they are mainly useful to store configuration values that do not change very often. The difference with flash is that individual bytes can be erased and that they usually have a longer lifespan.
```

aykevl (commented):
Looks good to me!

PR #319, opened 2023-01-04
Fix link to DFU Description.

trichner:
The trailing ) broke the link, added a space to help markdown a bit.

### DIFF
--- a/content/docs/reference/microcontrollers/feather-stm32f405.md
+++ b/content/docs/reference/microcontrollers/feather-stm32f405.md
@@ -67,7 +67,7 @@ The [Adafruit Feather STM32F405](https://www.adafruit.com/product/4382) is a tin
 
 ## Flashing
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Perhaps rewording this would be better, e.g.

> Flashing this board using its DFU bootloader can be a bit cumbersome, but it is possible - without requiring an external programmer - by pulling the B0 pin high (you can use the board's 3.3V output pin) at bootup when connected to host PC via USB. This puts the device in bootloader mode.
> For more information see Adafruit docs at https://learn.adafruit.com/adafruit-stm32f405-feather-express/dfu-bootloader-details
```

trichner:
> Perhaps rewording this would be better, e.g.
> 
> > Flashing this board using its DFU bootloader can be a bit cumbersome, but it is possible - without requiring an external programmer - by pulling the B0 pin high (you can use the board's 3.3V output pin) at bootup when connected to host PC via USB. This puts the device in bootloader mode.
> > For more information see Adafruit docs at https://learn.adafruit.com/adafruit-stm32f405-feather-express/dfu-bootloader-details

fair point, I now took your suggestion and actually linked the text instead of showing the entire ugly URL (consistent what exists further above in the document).

trichner:
Any interest in getting it fixed? I think I addressed all the feedback.

trichner:
Ping?

deadprogram:
Thanks for the reminder @trichner now merging.

PR #321, opened 2023-01-10
getting-started: remove `avr-gcc` dependency

aykevl:
This dependency is no longer needed to build binaries for AVR. `avrdude` is still needed however.


deadprogram:
:scissors: thanks @aykevl now merging.

PR #323, opened 2023-01-31
concept: low power

ysoldak:
Bits and pieces from different slack conversations put together.

### DIFF
--- a/content/docs/concepts/low-power.md
+++ b/content/docs/concepts/low-power.md
@@ -1,0 +1,74 @@
+---
+title: "Low Power"
+linkTitle: "Low Power"
+type: "docs"
+weight: 35
+description: >
+  How to consume less power on microcontrollers
+---
+
+Microcontrollers are usually rather non-demanding when it comes to power consumption.
+Couple hudreds mW is the most you may expect a hobby project would use.
+
+Sometimes, even that is not low enough.
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and peripherals continue to consume power while your program "sleeps".
+
+The standard way to pause execution in Go is to use `time.Sleep()` function that is available in TinyGo too. Depending on the chip and effort put into its support, the efficiency of this function may vary. The only thing certain: this is _not worse_ than busy looping until a certain time has passed.  
+Still, with such high-level function as `time.Sleep()` it is not always possible to disable everything that consumes power as the decisions may depend on the application, presense of peripherals and the board used.
+
+Most microcontrollers have one or more "low power" modes, also known as "deep sleep" and, sometimes, "dormant".  
+In such modes microcontroller may consume as low as milli- or even micro-Watts.  
+TinyGo gives you low level access to hardware to try and make most of it.
+
+## Implementations
+
+Manufacturers implement different modes, depending of capabilites of their products and name these modes differently too.
+Hence no naming consistency across microcontrollers, unfortunately.
+
+Always consult "Power consumption" part of a respective datasheet to understand capabilities of your hardware.  
+When reading datasheets, keep in mind peripherals that draw power too. These include \[power\] LEDs, various sensors or co-processors usually present on development boards.
+
+Capabilities of different chips:
+- Some very good chips (like the nRF series) can go to a very low power state (just a few µA) with RAM retention and waking up from a low-power RTC.
+- Some limited chips (like the esp32) can't really "sleep" with low power -- they can disable the core and memory and reset after a predefined time losing state stored in memory.
+- Other chips need to use a special feature to go into deep sleep mode.
+
+Especially the distinction between waking up with or without RAM is important, because it greatly affects how you write software.
+
+The common way to wake up from deep sleep is to use [interrupts](compiler-internals/interrupts), such as timer, real-time-clock or an interrupt on a pin change.
+
+### RP2040
+
+Minimal achievable power consumption is 0.18mA in "Dormant" and 0.39mA in "Sleep" modes.  
+See [RP2040 datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf#page=621), section 5.7 Power Consumption.
+
+### nRF Family
+
+[nRF series chips](https://www.nordicsemi.com/Products) are exceptionally efficient and is recommended choice for low-power projects.  
+Even with [BLE broadcasting enabled](https://devzone.nordicsemi.com/power/w/opp/2/online-power-profiler-for-bluetooth-le) it can use less power than RP2040 in its lowest power state!
+
+Minimal achievable power consumption is in the area of µA.
+
+TinyGo implements standard `time.Sleep()` very efficiently for nRF chips so you can just use it and have great results.  
+The only thing you may want to tweak is to disable the UART with help of `-serial=none` flag to go even lower than default.
+
+## The Way to Low Power
+
+> Attached [debugger](../guides/debugging) usually blocks the low power state in chip.  
+> The best way to measure when using a debugger is to flash the code, do a power cycle, and then measure.
+
+> The changes discussed in this paragraph shall be made in TinyGo and not in your application code.  
+> These are `runtime` and/or `machine` packages, in files related to your target.
+
+Following steps shall help you achieve low power consumption:
+
+- Put a `for { arm.Asm("wfi") }` loop before anything else is initialized (at the beginning of [Reset_Handler](https://github.com/tinygo-org/tinygo/search?q=%22export+Reset_Handler%22)).  
+  _`wfi` pauses the processor, puts it in "wait for interrupt" state. And when inserted before any initialisation, this ensures no internal subsystem is running consuming power._
+- Measure the current, this should be the lowest consumption as advertised for the chip.
+- If it is higher than that, investigate why (it could be a power LED, regulator, some clock that's enabled by the bootloader, maybe other things).  
+  _Usually this ranges somewhere between 1µA and 100µA, depending on the low-power capabilities of the chip. Don't continue until you really do reach the lowest possible power consumption._
+- Slowly move the loop to later in the code, until you move it into the main function.
+- There are probably things initialized by default that consume current, like UART, PLLs, different clocks. Disable them before continuing, or investigate why they consume power.


ysoldak:
See also https://github.com/tinygo-org/tinygo/issues/3014

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,64 @@
...
+
+The things needed to make nrf series chips efficient are:
+- Disable the UART with help of `-serial=none` flag;
+- Use the special sleep function of the SoftDevice;
+
+## The Way to Low Power
+
+Following steps shall help you achieve low power consumption:
+
+- Put a for `{ arm.Asm("wfi") }` loop before anything else is initialized (at the beginning of `Reset_Handler`).
```
soypat: Couple comments- `for` should be inside the code tildes. And maybe an explanation of what  `Reset_Handler` is? I've never used it.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,64 @@
...
+Couple of watts is the most you may expect a hobby project would use.
+
+Sometimes, that is not low enough.
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and periferals consume power as usual while your program "sleeps".
+
+To address that, most microcontrollers have one or more "low power" modes, also known as "deep sleep" and, sometimes, "dormant".  
+In such modes microcontroller may consume as low as milli- or even micro-Watts.
```
soypat, line 23: Maybe this would be a good place to make the distinction between `time.Sleep` and these "deeper" sleeps? i.e:

When writing normal Go code one has the standard library `time.Sleep` function at disposal. While this sleep saves us power and is more efficient than looping until a certain time has passed, there are other ways of letting the CPU sleep. When one has baremetal control of the hardware, as is the case when writing TinyGo for flashing to microcontroller chips, it is possible to put several critical parts of a micocontroller to sleep, and sometimes put the uC to sleep entirely to save power. 

Text is for demo. has not been reviewed closely

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,64 @@
...
+
+## The Way to Low Power
+
+Following steps shall help you achieve low power consumption:
+
+- Put a for `{ arm.Asm("wfi") }` loop before anything else is initialized (at the beginning of `Reset_Handler`).
+- Measure the current, this should be the lowest consumption as advertised for the chip.
+- If it is higher than that, investigate why (it could be a power LED, regulator, some clock that's enabled by the bootloader, maybe other things).  
+  Usually this ranges somewhere between 1µA and 100µA, depending on the low-power capabilities of the chip. Don't continue until you really do reach the lowest possible power consumption.
+- Slowly move the loop to later in the code, until you move it into the main function.
```
eyelight, line 73: Might be helpful for the benefit of casual hobbyists to discuss (either here or line 59) what the loop is doing, and what "wfi" means (wait for instruction?). 

Also, where to look for the functions that will be run 'before' main(). My best guess is that you're referencing `src/runtime/runtime_[chip].go` and testing the infinite 'wfi' loop before each call in the `init` function. (I will definitely try this!)

content/docs/concepts/low-power.md
ysoldak: `Reset_Handler` is the "main's main" :D I don't know how to describe it best, probably just put the search link under the word?
https://github.com/tinygo-org/tinygo/search?q=%22export+Reset_Handler%22

ysoldak:
```suggestion
- Put a `for { arm.Asm("wfi") }` loop before anything else is initialized (at the beginning of [Reset_Handler](https://github.com/tinygo-org/tinygo/search?q=%22export+Reset_Handler%22)).  
  `wfi` pauses the processor, puts it in "wait for interrupt" state. And when inserted before any initialisation, this ensures no internal subsystem is running consuming power.
```

content/docs/concepts/low-power.md
ysoldak: Yes, you are right, such `init` function you can find by search for "export Reset_Handler" in tinygo repository and picking your architecture. I've added change suggestion on line 59, shall be easier to grasp now.

"wfi" pauses processor in "wait for interrupt", that basically makes processor "do nothing". And if inserted before any initialisation, we can be sure no any internal subsystem is running consuming power.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,64 @@
...
+
+Microcontrollers are usually rather non-demanding when it comes to power consumption.
+Couple of watts is the most you may expect a hobby project would use.
+
+Sometimes, that is not low enough.
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and periferals consume power as usual while your program "sleeps".
+
```
ysoldak, line 18:
```suggestion

The standard way to pause execution is to use `time.Sleep()` function that is available in TinyGo too. Depending on the microcontroller architecture and effort put into it, the efficiency of this function can vary. The only thing is certain: it is better than busy looping until a certain time has passed.  
Still, with such high-level function as `time.Sleep()` it is not always possible to disable everything that consumes power as the decisions may depend on the application or the board used.

```

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,64 @@
...
+To address that, most microcontrollers have one or more "low power" modes, also known as "deep sleep" and, sometimes, "dormant".
```
ysoldak:
```suggestion
Most microcontrollers have one or more "low power" modes, also known as "deep sleep" and, sometimes, "dormant".  
```

content/docs/concepts/low-power.md
ysoldak, line 23:
```suggestion
In such modes microcontroller may consume as low as milli- or even micro-Watts.  
TinyGo gives you low level access to hardware to try and make most of it.
```

ysoldak: Thank you, @soypat, changes applied

ysoldak:
@soypat @eyelight PTAL.
@deadprogram @aykevl any ideas how to improve this page?

How do we work with docs in general?
Shall a page be acceptable as long as it does not confuse users and does not contain outright lies?

eyelight (approved):

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+Microcontrollers are usually rather non-demanding when it comes to power consumption.
+Couple of watts is the most you may expect a hobby project would use.
+
+Sometimes, that is not low enough.
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and periferals consume power as usual while your program "sleeps".
+
+The standard way to pause execution is to use `time.Sleep()` function that is available in TinyGo too. Depending on the microcontroller architecture and effort put into it, the efficiency of this function can vary. The only thing is certain: it is better than busy looping until a certain time has passed.
```
ysoldak:
```suggestion
The standard way to pause execution in Go is to use `time.Sleep()` function that is available in TinyGo too. Depending on the microcontroller architecture and effort put into it, the efficiency of this function can vary. The only thing is certain: it is better than busy looping until a certain time has passed.  
```

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+type: "docs"
+weight: 35
+description: >
+  How to consume less power on microcontrollers
+---
+
+Microcontrollers are usually rather non-demanding when it comes to power consumption.
+Couple of watts is the most you may expect a hobby project would use.
+
+Sometimes, that is not low enough.
```
ysoldak:
```suggestion
Sometimes, even that is not low enough.
```

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and periferals consume power as usual while your program "sleeps".
```
ysoldak:
```suggestion
The main caveat here is if not configured properly, microcontroller and peripherals continue to consume power as usual while your program "sleeps".
```

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+Manufacturers implement different modes, depending of capabilites of their products and name these modes differently too.
+Hence no naming consistency across microcontrollers, unfortunately.
+
+Always consult "Power consumption" part of a respective datasheet to understand capabilities of your hardware.  
+When reading datasheets, keep in mind peripherals that draw power too. These include \[power\] LEDs, various sensors or co-processors usually present on development boards.
+
+Capabilities of different chips:
+- Some very good chips (like the nRF series) can go to a very low power state (just a few µA) with RAM retention and waking up from a low-power RTC.
+- Some limited chips (like the esp32) can't really "sleep" with low power -- they can disable the core and memory and reset after a predefined time.
```
ysoldak:
```suggestion
- Some limited chips (like the esp32) can't really "sleep" with low power -- they can disable the core and memory and reset after a predefined time losing state stored in memory.
```

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+The things needed to make nrf series chips efficient are:
+- Disable the UART with help of `-serial=none` flag;
+- Use the special sleep function of the SoftDevice;
+
+## The Way to Low Power
+
+Following steps shall help you achieve low power consumption:
+
+- Put a `for { arm.Asm("wfi") }` loop before anything else is initialized (at the beginning of [Reset_Handler](https://github.com/tinygo-org/tinygo/search?q=%22export+Reset_Handler%22)).
```
soypat, line 68: That loop is infinite? Should there be a `isInitialized()` in the for loop clause?

soypat (approved):
I've left one last comment but this looks good as it is!

content/docs/concepts/low-power.md
ysoldak: As I understood @aykevl idea, this loop shall permanently lock the core in some specific place. And nothing need to be initialised, the asm call is fine as it is.
Then you move this lock down the code to find the place system starts eating more, and address that.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+title: "Low Power"
+linkTitle: "Low Power"
+type: "docs"
+weight: 35
+description: >
+  How to consume less power on microcontrollers
+---
+
+Microcontrollers are usually rather non-demanding when it comes to power consumption.
+Couple of watts is the most you may expect a hobby project would use.
```
aykevl: I think milliwatts are more realistic, usually microcontrollers don't use more than 10mA or so, 100mA in peaks maybe.

Also, perhaps amperes (milliamperes) are more useful here? When I see power consumption in datasheets and such, it is usually measured in milliamperes not in watts (their voltage is constant, or the current consumption even goes _up_ on a higher voltage - in that case current is more precise).

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+Sometimes, even that is not low enough.
+Power your project from a battery, and it'd benefit greatly from consuming less.
+
+Common scenario for such projects is sleeping most of the time, and only rarely briefly wake up to query sensors and execute actions.
+The main caveat here is if not configured properly, microcontroller and peripherals continue to consume power as usual while your program "sleeps".
+
+The standard way to pause execution in Go is to use `time.Sleep()` function that is available in TinyGo too. Depending on the microcontroller architecture and effort put into it, the efficiency of this function can vary. The only thing certain: this is better than busy looping until a certain time has passed.
```
aykevl: > The only thing certain: this is better than busy looping until a certain time has passed.

I'm afraid not even that is true. Sometimes, `time.Sleep` is implemented as a busy loop until a given time has passed as a stopgap until proper sleeping has been implemented. But it's unlikely to be _worse_ at least.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+Always consult "Power consumption" part of a respective datasheet to understand capabilities of your hardware.  
+When reading datasheets, keep in mind peripherals that draw power too. These include \[power\] LEDs, various sensors or co-processors usually present on development boards.
+
+Capabilities of different chips:
+- Some very good chips (like the nRF series) can go to a very low power state (just a few µA) with RAM retention and waking up from a low-power RTC.
+- Some limited chips (like the esp32) can't really "sleep" with low power -- they can disable the core and memory and reset after a predefined time losing state stored in memory.
+- Other chips need to use a special feature to go into deep sleep mode.
+
+Especially the distinction between waking up with or without RAM is important, because it greatly affects how you write software.
```
aykevl, line 39: :+1:

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+The common way to wake up from deep sleep is to use [interrupts](compiler-internals/interrupts), such as timer, real-time-clock or an interrupt on a pin change.
+
+### RP2040
+
+Minimal achievable power consumption according to [RP2040 datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf): 6mW to 12mW (0.19mA at 3.3V)
```
aykevl: It would be nice to link to a specific page in the datasheet (add `#page=xxx` to the end of the URL) and/or also specify the section where this is specified.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+[nRF series chips](https://www.nordicsemi.com/Products) are exceptionally efficient and is recommended choice for low-power projects.
+
+Minimal achievable power consumption is in the area of micro-Watts.
+
+Even with BLE broadcasting enabled it can be lower power than rp2040 in its lowest power state!
+
+The things needed to make nrf series chips efficient are:
+- Disable the UART with help of `-serial=none` flag;
+- Use the special sleep function of the SoftDevice;
```
aykevl: > - Use the special sleep function of the SoftDevice;

This is implemented in `time.Sleep` so I don't think it's needed to mention this. See: https://github.com/tinygo-org/tinygo/blob/release/src/runtime/runtime_nrf_softdevice.go#L17

My low-power project is here: https://github.com/aykevl/things/tree/master/poi
The only special thing I do is `-serial=none`, other than that all the regular things work just fine (`time.Sleep` enters low power state, and BLE just works with low power when using tinygo.org/x/bluetooth!).

For nRF (nrf52xxx specifically) it may also be useful to link to https://devzone.nordicsemi.com/power/w/opp/2/online-power-profiler-for-bluetooth-le, which calculates how much current these devices consume when correctly used (note that this page, like most others, uses mA and µA instead of watts).

content/docs/concepts/low-power.md
aykevl: Yes, it's just for testing. It's a way to make sure you start from a place where nothing is initialized.

Oh, now I remember something else: when a debugger is attached, that will usually block the low power state. So the best way to measure is to flash the code, then do a power cycle, and then measure. Probably worth adding to the documentation.

content/docs/concepts/low-power.md
```diff
@@ -0,0 +1,69 @@
...
+
+## The Way to Low Power
+
+Following steps shall help you achieve low power consumption:
```
aykevl: Perhaps worth pointing out this is needed if you want to add a low power state to your particular chip? That is, it's not really an application-level change but a change to TinyGo (runtime/machine package).

aykevl (commented):
Looks reasonable to me!
Ideally this document would be very short and only list the quirks of each particular device (for example, for nRF devices to disable UART). But until we have full support it's better to just document everything.

content/docs/concepts/low-power.md
ysoldak: I was more thinking about a hobby project as a whole, with sensors, leds, relays, servos, etc.
But yeah, we can be less dramatic here. "couple hundreds mA" shall be fine, right?

content/docs/concepts/low-power.md
ysoldak: I'll try to re-formulate, thanks!

content/docs/concepts/low-power.md
ysoldak: Citing you here :D

content/docs/concepts/low-power.md
ysoldak: Good, will do.

content/docs/concepts/low-power.md
ysoldak: As for watts vs. amps... if you say amps -- the next question is: how many volts?
Also, people are used to watts, in light bulbs, in heating elements, on laptop chargers. And at the end of a month, when it's time to see kWh (oh my god).
But yeah, batteries are in mAh and we are measuring amps. So probably I mention watts in parentheses first time and then switch to amps entirely.

content/docs/concepts/low-power.md
ysoldak: I don't know... access to registers and all is actually available in application code. So people can do it in their projects directly and only after that contribute to TinyGo, perhaps...

content/docs/concepts/low-power.md
ysoldak: Added as a link under "BLE broadcasting enabled" in above para.

content/docs/concepts/low-power.md
ysoldak: Added about the debugger, thanks.

content/docs/concepts/low-power.md
ysoldak: Removed watts from everywhere but the second sentence. It starts now "Couple hudreds mW"

content/docs/concepts/low-power.md
ysoldak: Added a note about the changes have to be done in TinyGo, not app code.

ysoldak:
I hope all comments were addressed, @aykevl

deadprogram:
Thanks everyone! Merging.

PR #324, opened 2023-02-12
install/linux: fully remove avr-gcc dependency

aykevl:
Looks like I missed some.


deadprogram:
Thanks @aykevl now merging.

PR #327, opened 2023-03-05
docs/reference: add Go compatibility matrix

dImrich:
The compatibility matrix which shows which versions of TinyGo are compatible with which versions of Go.

Not all versions of Go are compatible with all versions of TinyGo. Therefore, having a compatibility matrix can be helpful in determining which version of TinyGo is best suited for a specific version of Go.


aykevl:
This looks reasonable, but who is going to maintain this? We need to maintain this information for it to be useful.

dImrich:
Yeah you are absolutely right, this also occured to me shortly after I submitted this PR. Maintaining this would be utter pain in the ass.

Maybe I can make some script that we can add somewhere to this repo which would parse condition in `tinygo-org/tinygo/builder/config.go#L36` and add new line to this table if needed. But it's not the best solution to this maintability problem as `builder/config.go` could change in dramatic way not so long down the road. 

I should really just close this. Support matrix like this would save me a lot of time but that's only my small use case, most people will probably just go with newest versions of tinygo and golang and it will work out of the box.

aykevl:
Actually, on second thought I think it would be a good idea.

aykevl:
Also see: https://github.com/tinygo-org/tinygo/pull/3682

PR #328, opened 2023-03-06
content/docs: Fix broken links

bxparks:
I wanted to fix one broken link. Then saw a bunch of others, so decided to fix all the ones that I could find.
Verified by running a local `hugo serve` and manually checking that the links are fixed.

### DIFF
--- a/content/docs/guides/tips-n-tricks.md
+++ b/content/docs/guides/tips-n-tricks.md
@@ -66,4 +66,4 @@ buf2 := t.buf[2:6]
 
 For more recipes on how to avoid or minimize allocations while working with slices, please see [additional tricks](https://github.com/golang/go/wiki/SliceTricks#additional-tricks) section of [slice tricks](https://github.com/golang/go/wiki/SliceTricks) page.
 
-Familiarize yourself with [heap allocation concept](../concepts/compiler-internals/heap-allocation.md) and use it to your advantage.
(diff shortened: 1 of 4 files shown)


content/docs/guides/tips-n-tricks.md
```diff
@@ -66,4 +66,4 @@ buf2 := t.buf[2:6]
 
 For more recipes on how to avoid or minimize allocations while working with slices, please see [additional tricks](https://github.com/golang/go/wiki/SliceTricks#additional-tricks) section of [slice tricks](https://github.com/golang/go/wiki/SliceTricks) page.
 
-Familiarize yourself with [heap allocation concept](../concepts/compiler-internals/heap-allocation.md) and use it to your advantage.
+Familiarize yourself with [heap allocation concept]({{<ref "../concepts/compiler-internals/heap-allocation.md">}}) and use it to your advantage.
```
scottfeldman, line 69: I think this link is still broken.  It should be:

../../concepts/compiler-internals/heap-allocation/

bxparks: The Hugo generator seems to canonicalize those links. See https://gohugo.io/functions/ref/
I double-checked all my fixes with a local hugo instance, so I think mine are correct.
It's possible that I may have *missed* some broken links though.

bxparks: A better Hugo link here: https://gohugo.io/content-management/shortcodes/#ref-and-relref

aykevl (approved):
Looks reasonable.

PR #329, opened 2023-03-07
docs/tutorial: Add serial monitor tutorial

bxparks:
In 4 subsections:

1) Writing to the serial port using fmt.Print(), fmt.Println(), and the
   built-in functions print() and println(). Explain flash memory consumption
   of `fmt` package.
2) Running `tinygo monitor` to see the output on the host computer. 
3) Reading from the serial port.
4) Alternative serial monitors (Arduino IDE, pyserial, picocom).

### DIFF
--- a/content/docs/tutorials/serialmonitor.md
+++ b/content/docs/tutorials/serialmonitor.md
@@ -132,6 +132,16 @@ In this example, the serial monitor missed the first 4 lines of "Hello, World"
 (0 to 3) because the program started to print those lines immediately after
 flashing, but before the serial monitor was connected.
 
+On some AVR microcontrollers, the speed of the serial port is set to 9600
+instead of the default 115200. So you need to override the baud rate of `tinygo
+monitor` like this:
+```
+$ tinygo monitor -baudrate=9600
+```
+
+See [Alternative Serial Monitors](#alternative-serial-monitors) for an
+explanation of the "baud rate".
+
 ## Serial Input
 
 Occasionally it is useful to send characters from the host computer to the
@@ -139,7 +149,7 @@ microcontroller. The following program reads a single byte from the
 `machine.Serial` object and prints the character back to the host computer.
 
 The caveat is that the `Serial.ReadByte()` feature is *not* currently
-implemented on every microcontrollers supported by TinyGo. For example, the
+implemented on every microcontroller supported by TinyGo. For example, the
 following program does not work on the ESP32 or the ESP8266.
 
 ```go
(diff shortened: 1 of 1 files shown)


content/docs/tutorials/serialmonitor.md
```diff
@@ -0,0 +1,294 @@
...
+
+* `/dev/ttyUSB0`
+* `/dev/ttyACM0`
+
+On MacOS machines, the serial port will look like this:
+
+* `/dev/cu.usbserial-1420`
+* `/dev/cu.usbmodem6D8733AC53571`
+
+[TODO: No idea on Windows.]
```
sago35: On Windows, strings such as `COM1` and `COM31` are used as port names.

sago35 (commented):
Thank you very much.
I wrote information about port names on Windows.

It would also be nice to note that some Arduino and other things that use UARTtoUSB chips use speeds such as 9600bps.

ex) `tinygo monitor --target arduino --baudrate 9600`

The following can also use tinygo flash -monitor.

```
$ tinygo flash -target=arduino-zero
$ tinygo monitor
```

```
$ tinygo flash -target=arduino-zero -monitor
```

content/docs/tutorials/serialmonitor.md
bxparks: Added `COM1` and `COM31`. Thanks.

I didn't know about the default being 9600 on AVR processors. 
* Added `tinygo monitor -baudrate=9600` for AVR.
* Added paragraph explaining which processors use 115200 and which use 9600.

Is there a reason why 115200 doesn't work on AVR? When I program them with the Arduino IDE/CLI, I always use 115200 for the serial monitor, and I haven't run into any issues.

Fixed a couple of grammar typos.

I am reluctant to add the `-monitor` flag in this tutorial. Because it does not work consistently. The monitor starts too quickly after flashing, and it terminates with an error message like this:
```
$ tinygo flash -target=arduino-zero -monitor
[...]
Verify successful
Done in 0.483 seconds
Connected to /dev/ttyACM0. Press Ctrl-C to exit.
error: read error: Port has been closed
```
The `-monitor` flag seems to work only for microcontroller boards using a USB-to-serial adapter, which presumably keeps the serial port connection alive during the flashing and reboot process.

I do mention the `-monitor` flag in the subcommand reference that I added in PR #330.

sago35: The AVR does not necessarily need to be 9600 bps.
However, since TinyGo defaults to 9600 bps, I believe the `-baudrate` flag needs to be explained.

https://github.com/tinygo-org/tinygo/blob/v0.27.0/src/machine/machine_atmega.go#L161-L165

sago35: > I am reluctant to add the -monitor flag in this tutorial. Because it does not work consistently. The monitor starts too quickly after flashing, and it terminates with an error message like this:

I added `tinygo monitor` and I am aware that it does not work with some targets.
However, it works fine on my samd21/samd51/rp2040/nrf52840.

* samd21 : xiao / feather-m0
* samd51 : feather-m4 / wioterminal
* rp2040 : pico / xiao-rp2040
* nrf52840 : feather-nrf52840 / xiao-ble

And it works better with USBtoSerial.

I believe the case of failure as in your example is the case of a slow reboot of the microcontroller after flashing.
`tinygo monitor` accidentally connects to bootloader and disconnects at reboot after flash.
If you are using `UF2 bootloader`, changing to a newer version should solve many of your problems.

https://github.com/microsoft/uf2-samdx1

bxparks: Please take a look at the latest version. I think I incorporated most (if not all) of your comments. The behavior of `tinygo monitor` when there are multiple boards attached is not consistent. Sometimes it can figure out the port, sometimes it cannot, and sometimes `flash` will figure it out, but `-monitor` does not. I tried to explain this without overwhelming the beginner...

sago35: Thank you very much. I think it is a very good and easy to understand document.

sago35 (approved):
LGTM
Thank you very much.
I think it is a very good and easy to understand document.

deadprogram:
Thank you @bxparks now merging.

PR #334, opened 2023-03-10
microcontrollers/_index.md: well supported boards

bxparks:
Add a list of "well supported" microcontrollers (SAMD21, SAMD51, nRF52840, RP2040), and a short list of popular, but not well-supported, microcontrollers (AVR, ESP8266, ESP32). I did not add any recommendations of specific brands (e.g. Adafruit, Sparkfun, Seeed, Wemos, Raspberry, Arduino, etc.) because I think that list would be subjective, and probably needs to be different based on the user's place of residence.

### DIFF
--- a/content/docs/reference/microcontrollers/_index.md
+++ b/content/docs/reference/microcontrollers/_index.md
@@ -10,6 +10,37 @@ TinyGo lets you run Go directly on microcontrollers.
 
 TinyGo has support for 85 different boards and devices such as the Arduino Nano33 IoT, Adafruit Circuit Playground Express, BBC micro:bit and more. Click on a board name below to see the what features are supported for the given hardware.
 
+Support for some boards and processor types are more complete than others.
+As of early 2023, boards using the following microcontrollers are
+well-supported:
+
+* [SAMD21](https://www.microchip.com/en-us/product/ATSAMD21G18) based on the
+  ARM Cortex-M0+ processor
+* [SAMD51](https://www.microchip.com/en-us/product/ATSAMD51N19A) based on the
+  ARM Cortex-M4 processor
+* [nRF52840](https://infocenter.nordicsemi.com/topic/ps_nrf52840/keyfeatures_html5.html)
+  based on the Arm Cortex-M4F processor
+* [RP2040](https://en.wikipedia.org/wiki/RP2040) with dual ARM Cortex-M0+
+  processors
+
+The introductory Arduino boards based on the 8-bit AVR processors work
+relatively well under TinyGo. But they have limited amounts of flash and static
+memory so they support only small applications:
+
+* [ATmega328P](https://www.microchip.com/en-us/product/ATmega328P), used by
+  Arduino Nano, Arduino UNO, etc.
+
+Boards using the Espressif microcontrollers have become popular in IoT
+applications because of their support for WiFi. Unfortunately TinyGo does not
+support WiFi nor Bluetooth on these boards:
+
+* [ESP8266](https://en.wikipedia.org/wiki/ESP8266) based on the Xtensa LX106
+  processor
+* [ESP32](https://en.wikipedia.org/wiki/ESP32) based on the Xtensa LX6
+  processor
+* [ESP32-C3](https://www.espressif.com/en/products/socs/esp32-c3) based on the
+  RISC-V processor
+
 We also give you the ability to add new boards. If your target isn't listed here, please raise an issue in the [issue tracker](https://github.com/tinygo-org/tinygo/issues).
 
 Want to know the details about how it is possible to compile Go for microcontrollers? Check out the [microcontrollers](../../concepts/compiler-internals/microcontrollers/) page in our "Compiler Internals" section.
--- a/content/docs/reference/microcontrollers/_index.md
+++ b/content/docs/reference/microcontrollers/_index.md
@@ -16,16 +16,27 @@ well-supported:
 
 * [SAMD21](https://www.microchip.com/en-us/product/ATSAMD21G18) based on the
   ARM Cortex-M0+ processor
+    * Some companies (Adafruit) call these boards the "M0".
 * [SAMD51](https://www.microchip.com/en-us/product/ATSAMD51N19A) based on the
   ARM Cortex-M4 processor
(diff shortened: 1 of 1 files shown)


content/docs/reference/microcontrollers/_index.md
```diff
@@ -10,6 +10,37 @@ TinyGo lets you run Go directly on microcontrollers.
...
+Support for some boards and processor types are more complete than others.
+As of early 2023, boards using the following microcontrollers are
+well-supported:
+
+* [SAMD21](https://www.microchip.com/en-us/product/ATSAMD21G18) based on the
+  ARM Cortex-M0+ processor
+* [SAMD51](https://www.microchip.com/en-us/product/ATSAMD51N19A) based on the
+  ARM Cortex-M4 processor
+* [nRF52840](https://infocenter.nordicsemi.com/topic/ps_nrf52840/keyfeatures_html5.html)
+  based on the Arm Cortex-M4F processor
```
aykevl: Most nRF5x boards are well supported:
* nrf52832 also has good support (but is less commonly used by Adafruit and the like)
* nrf52833 is used in the microbit v2, so it probably also works fine
* nrf51822 is older, has less memory, but I expect it also works fine
* there may be other nrf52xxx chips that may not be as well supported, but that need only small tweaks to work well

content/docs/reference/microcontrollers/_index.md
```diff
@@ -10,6 +10,37 @@ TinyGo lets you run Go directly on microcontrollers.
...
+* [RP2040](https://en.wikipedia.org/wiki/RP2040) with dual ARM Cortex-M0+
+  processors
```
aykevl: It's correct that it's dual core, but saying it here may lead people to expect that TinyGo is also able to use both cores (which it isn't). So I'd recommend leaving that out.

content/docs/reference/microcontrollers/_index.md
```diff
@@ -10,6 +10,37 @@ TinyGo lets you run Go directly on microcontrollers.
...
+
+The introductory Arduino boards based on the 8-bit AVR processors work
+relatively well under TinyGo. But they have limited amounts of flash and static
+memory so they support only small applications:
```
aykevl: True. Perhaps giving a few specific examples might help? For example: you can't use `fmt` on these chips and goroutines don't work so well due to limited RAM.

content/docs/reference/microcontrollers/_index.md
```diff
@@ -10,6 +10,37 @@ TinyGo lets you run Go directly on microcontrollers.
...
 TinyGo has support for 85 different boards and devices such as the Arduino Nano33 IoT, Adafruit Circuit Playground Express, BBC micro:bit and more. Click on a board name below to see the what features are supported for the given hardware.
 
+Support for some boards and processor types are more complete than others.
+As of early 2023, boards using the following microcontrollers are
+well-supported:
+
+* [SAMD21](https://www.microchip.com/en-us/product/ATSAMD21G18) based on the
+  ARM Cortex-M0+ processor
+* [SAMD51](https://www.microchip.com/en-us/product/ATSAMD51N19A) based on the
+  ARM Cortex-M4 processor
```
aykevl: Perhaps it is useful to give their branding names. The SAMD21 chips are called "M0" by Adafruit and "Zero" by Arduino. The SAMD51 board is called "M4" by Adafruit (Arduino doesn't seem to be selling any SAMD51 boards).

aykevl (commented):
This is great! A few suggestions below.

bgould:
I think an important detail for newcomers might that some boards support loading code over USB vs needing a separate programmer.  Similarly, some boards/chips support serial-over-USB without any extra hardware and some will require a separate UART-USB converter.  I assume that most newcomers might prefer for things to "just work" out of the box over a USB connection.

content/docs/reference/microcontrollers/_index.md
bxparks: Added "other nRF" boards with references to nrf52833,  nrf52832, nrf51822. I'm not familiar with any of these, and they seem to be rare under `tinygo/targets`:

* grep -i nrf51822 * | wc: 1
* grep -i nrf52832 * | wc: 1
* grep -i nrf52833 * | wc: 8
* grep -i nrf52840 * | wc: 39

content/docs/reference/microcontrollers/_index.md
bxparks: I think the RP2040 is famous for being a cheap dual core, so people will want to know that. But it's a bit of important information that TinyGo supports only a single core, so I added that as a parenthetical remark.

content/docs/reference/microcontrollers/_index.md
bxparks: > True. Perhaps giving a few specific examples might help? For example: you can't use `fmt` on these chips and goroutines don't work so well due to limited RAM.

Yup, added

content/docs/reference/microcontrollers/_index.md
bxparks: > Perhaps it is useful to give their branding names. The SAMD21 chips are called "M0" by Adafruit and "Zero" by Arduino. The SAMD51 board is called "M4" by Adafruit (Arduino doesn't seem to be selling any SAMD51 boards).

Done

bxparks:
>

bxparks:
Ugh, accidentally hit the wrong item in the drop down... reopening...

bxparks:
> I think an important detail for newcomers might that some boards support loading code over USB vs needing a separate programmer. Similarly, some boards/chips support serial-over-USB without any extra hardware and some will require a separate UART-USB converter. I assume that most newcomers might prefer for things to "just work" out of the box over a USB connection.

That is useful information, but I feel like that is answering a slightly different question. I'm trying to answer: "Which boards are well-supported by TinyGo?" But your information is more along the line of: "What boards should I buy or use?" There seems to be many more factors involved in that answer: cost, number of GPIO ports, I2C support, wifi/bluetooth support, flash memory, static memory, CPU speed, USB-to-serial adapter or USB-CDC, power consumption, physical size of dev board.

I feel like we should create a separate page for that.

bxparks:
Ping.. any other comments?
I think I addressed all the prior feedback, so I think this is ready to be merged.
Just want to clear off my pending PRs...

aykevl:
Yes, looks good to me.

PR #337, opened 2023-03-24
Add machine.CPUReset() doc

scottfeldman:
Add machine.SystemReset() to do a hard reset on device.  Only available for cortexm, currently.

### DIFF
--- a/content/docs/guides/build.md
+++ b/content/docs/guides/build.md
@@ -172,6 +172,12 @@ You have successfully built TinyGo from source. Congratulations!
 
 We're not done yet. Some extra things need to be built before you can start using TinyGo.
 
+If you haven't already, you need to download llvm-project (for the compiler-rt sources that are needed for most architectures):
+
+```shell
+make llvm-source
+```
+
 To be able to use TinyGo on a bare-metal target, you need to generate some files first:
 
 ```shell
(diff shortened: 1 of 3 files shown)


scottfeldman:
Reference https://github.com/tinygo-org/tinygo/pull/3595

aykevl:
As it happens, I was thinking we needed this function today :)

I'm not sure about the name `SystemReset`, it might lead people to think it will reset more than just the microcontroller (like external devices).

What about `CPUReset`, analogous to `CPUFrequency`? But that is maybe too specific, as a reset will normally also reset I2C, SPI, pin states, etc.

> Not all chips support SystemReset.

I don't think we need to say this, because practically all microcontrollers support resetting the chip.

scottfeldman:
> As it happens, I was thinking we needed this function today :)
> 
> I'm not sure about the name `SystemReset`, it might lead people to think it will reset more than just the microcontroller (like external devices).
> 
> What about `CPUReset`, analogous to `CPUFrequency`? But that is maybe too specific, as a reset will normally also reset I2C, SPI, pin states, etc.

CPUReset is good.  I'll update.

> > Not all chips support SystemReset.
> 
> I don't think we need to say this, because practically all microcontrollers support resetting the chip.

I don't have anything to test on other than contexm ARMs, so maybe we leave this comment and when someone updates it for the other cpus we can remove it?

deadprogram:
@scottfeldman can you please rebase this against `dev`? Thank you!

PR #339, opened 2023-04-01
reference/machine: add SPI documentation

aykevl:


content/docs/reference/machine.md
```diff
@@ -84,6 +84,63 @@ SetInterrupt sets an interrupt to be executed when the pin changes state. The pi
...
+    SDI       Pin
+    LSBFirst  bool
+    Mode      uint8
+}
+```
+
+The `SPIConfig` struct contains the configuration for the SPI peripheral.
+
+  * `Frequency` is the maximum frequency that will be used: for example `1 * MHz` for 1MHz. Depending on chip capabilities, this or a lower frequency will be selected. When not set (or set to 0), the default of 4MHz will be used.
+  * `SCK`, `SDO` and `SDI` are the clock, data out, and data in pins respectively, however support for setting pins other than the default pins may not be supported by a given SPI peripheral. Some chips are flexible and allow the use of any pin, while other boards only allow a limited range of pins or use fixed SCK/SDO/SDI pins. When these pins are left at the zero value, the default for the particular board is used.
```
bxparks, line 103: This is great documentation.

By coincidence, I spent a few days last week learning most of this on my own by writing a couple of device drivers. I discovered that:

1) Some SPI implementations (e.g. SAMD21, ESP32) require the `SDI` to be defined to the correct hardware ~~MOSI~~ MISO pin, even if ~~MOSI~~ MISO is not used by the hardware. (I spent a few hours tracking this down).

2) As far as I can tell, the default SPI pins for `SCK`, `SDO`, and `SDI` are used only if **all 3** parameters are left to their default value of `0`. As soon as one of them is set to something other than `0`, then the SPI defaults are not used. In other words, most implementations have this little code fragment:

```go
  if config.SCK == 0 && config.SDO == 0 && config.SDI == 0 {
    config.SCK = SPI0_SCK_PIN
    config.SDO = SPI0_SDO_PIN
    config.SDI = SPI0_SDI_PIN
  }
```

3) It's not clear to me if `machine.NoPin` is intended to be a valid value for the `SDO`, `SDI`, and `SCK` pins for all SPI implementations. I tried setting `SDI: machine.NoPin` for the `xiao` and `esp32` boards, but that did not work.

4) It wasn't until much later that I discovered that the `drivers.SPI` interface was already defined. It's a useful bit of information, but maybe it does not belong on this page.

deadprogram:
This is great, thanks @aykevl

content/docs/reference/machine.md
aykevl: > 1. Some SPI implementations (e.g. SAMD21, ESP32) require the `SDI` to be defined to the correct hardware ~MOSI~ MISO pin, even if ~MOSI~ MISO is not used by the hardware. (I spent a few hours tracking this down).

Yeah, that's not great. Sometimes it's a hardware limitation, other times it's just not implemented yet.

> 2. As far as I can tell, the default SPI pins for `SCK`, `SDO`, and `SDI` are used only if **all 3** parameters are left to their default value of `0`.

Correct. This is the case because `machine.Pin(0)` can be a valid pin. We could consider changing that so that `machine.Pin(0)` is not a valid pin (and equals `machine.NoPin`), but that unfortunately doesn't map as closely to the hardware as what we have now.


> 3\. It's not clear to me if `machine.NoPin` is intended to be a valid value for the `SDO`, `SDI`, and `SCK` pins for all SPI implementations.

Sometimes it's supported. Other times, it's either a hardware limitation or simply isn't implemented yet in the machine package.

PR #343, opened 2023-04-11
guides/build: add binaryen build step

aykevl:
This is necessary for WebAssembly.


deadprogram:
Thanks for the addition @aykevl now merging.

PR #346, opened 2023-04-25
guides/compatibility: suggest the new unsafe.Slice etc functions

aykevl:
Using `reflect.SliceHeader` and `reflect.StringHeader` has always been non-portable. So instead of suggesting workarounds, suggest the proper way these operations can be done.


deadprogram:
Thanks for this improvment @aykevl now merging.

PR #350, opened 2023-04-25
docs/guides: add pages for macOS and Windows

deadprogram:
This PR adds pages for macOS and Windows to the `docs/guides` section.

### DIFF
--- a/content/docs/guides/macos.md
+++ b/content/docs/guides/macos.md
@@ -1,0 +1,12 @@
+---
+title: "macOS support"
+weight: 4
+description: |
+  How to use TinyGo to create standard macOS executables.
+---
+
+Using TinyGo you can compile programs for macOS systems.
+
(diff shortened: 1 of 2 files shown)


aykevl:
This seems fine by me, but what about merging the three OS pages? Because although Linux is probably the one with the best support, from a user perspective all you need to know is that `GOOS` and `GOARCH` work the way you'd expect from the `go` toolchain.

deadprogram:
> what about merging the three OS pages?

I think we have some OS specific information that would prove useful, and this gives us a place to put it.

For example, explaining that https://github.com/aykevl/macos-minimal-sdk is used for macOS. Which perhaps should be under the TinyGo org?

aykevl:
> For example, explaining that https://github.com/aykevl/macos-minimal-sdk is used for macOS.

Yeah, maybe. It's not relevant to users in most cases, but when you are using CGo then it probably is.

aykevl:
> Which perhaps should be under the TinyGo org?

Yeah, I've transferred it just now.

PR #351, opened 2023-04-26
Add 'go fmt' step to avoid lint build failure.

inindev:
Document 'go fmt' step


aykevl:
Yes, seems reasonable.
If you use an IDE, it will typically do this for you so most people won't need this.

PR #352, opened 2023-04-27
docs/guides: additional resources for WebAssembly

deadprogram:
This PR adds more to docs/guides with additional resources for WebAssembly.

### DIFF
--- a/content/docs/guides/webassembly/resources.md
+++ b/content/docs/guides/webassembly/resources.md
@@ -1,0 +1,32 @@
+---
+title: "Additional Resources"
+weight: 4
+description: |
+  Additional Resources for using TinyGo with WASM/WASI
+---
+
+Here are a few resources for helping learn more about using TinyGo with WebAssembly.
+
+**Wasm By Example**
+https://wasmbyexample.dev/
+
+**Are We Wasm Yet ? - Part 1**
+https://elewis.dev/are-we-wasm-yet-part-1
+
+**Are We Wasm Yet ? - Part 2**
+https://elewis.dev/are-we-wasm-yet-part-2
+
+**Writing a WebAssembly Service in TinyGo for Wagi and Spin**
+https://www.fermyon.com/blog/tinygo-webassembly-favicon-server
+
+**wazero - TinyGo**
+https://wazero.io/languages/tinygo/
+
+**WasmEdge Runtime - Go**
+https://wasmedge.org/book/en/write_wasm/go.html
+
+**WASI and Node.js**
+https://k33g.hashnode.dev/series/wasi-nodejs
+
+**Wazero, first steps**
+https://k33g.hashnode.dev/series/wazero-first-steps


content/docs/guides/webassembly/resources.md
```diff
@@ -0,0 +1,32 @@
...
+https://wasmbyexample.dev/
+
+**Are We Wasm Yet ? - Part 1**
+https://elewis.dev/are-we-wasm-yet-part-1
+
+**Are We Wasm Yet ? - Part 2**
+https://elewis.dev/are-we-wasm-yet-part-2
+
+**Shrink Your TinyGo WebAssembly Modules by 60%**
+https://www.fermyon.com/blog/optimizing-tinygo-wasm
```
aykevl: This one still uses `-wasm-abi=generic` which has been removed a while ago. Otherwise it seems fine.

content/docs/guides/webassembly/resources.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Writing a WebAssembly Service in TinyGo for Wagi and Spin**
+https://www.fermyon.com/blog/tinygo-webassembly-favicon-server
+
+**wazero - TinyGo**
+https://wazero.io/languages/tinygo/
```
aykevl, line 23: This one contains some outdated information on `malloc` and `free`. I'm not sure we should encourage people reading this? (Or perhaps send a message to the blog post writers to update?)

content/docs/guides/webassembly/resources.md
```diff
@@ -0,0 +1,32 @@
...
+
+**WasmEdge Runtime - Go**
+https://wasmedge.org/book/en/write_wasm/go.html
+
+**An essay on the bi-directional exchange of strings between the Wasm module (with TinyGo) and Node.js (with WASI support)**
+https://www.wasm.builders/k33g_org/an-essay-on-the-bi-directional-exchange-of-strings-between-the-wasm-module-with-tinygo-and-nodejs-with-wasi-support-3i9h
```
aykevl: This post contains some incorrect examples. For example:

```go
//export hello 
func hello() *byte {
  return &(([]byte)("hello world")[0]) 
}
```

This code provides no way to prevent the returned object from being garbage collected. It should use `malloc` or at least store the value in a global.

aykevl (commented):
Good idea, but I'm not sure about some of the blog posts. If we're going to link to external posts, we should make sure they're at least using current recommendations and not all of them do unfortunately.

content/docs/guides/webassembly/resources.md
deadprogram: > Or perhaps send a message to the blog post writers to update?

That for sure! cc @codefromthecrypt

content/docs/guides/webassembly/resources.md
deadprogram: cc @k33g

content/docs/guides/webassembly/resources.md
deadprogram: cc @technosophos

content/docs/guides/webassembly/resources.md
k33g: @deadprogram it's an old blog post; this one is perhaps better https://k33g.hashnode.dev/wasi-communication-between-nodejs-and-wasm-modules-with-the-wasm-buffer-memory#heading-calling-a-wasm-function-from-nodejs-with-strings

I am writing several series:
- https://k33g.hashnode.dev/series/wasi-nodejs
- https://k33g.hashnode.dev/series/wazero-first-steps

deadprogram: Changed out for those new series which are :fire: great work @k33g thank you!!!

content/docs/guides/webassembly/resources.md
codefromthecrypt: the jury is still out on the malloc/free stuff, but coming in soon. We need to see benchmark before after and it is nearly there.

codefromthecrypt: basically there are a lot of people using the existing approach (which was reviewed by several) we owe it when doing an update to say what the new is and what it will cost you https://github.com/tetratelabs/wazero/pull/1390

aykevl: @codefromthecrypt I'm not sure what you mean? `malloc` and `free` has been safe for use since https://github.com/tinygo-org/tinygo/pull/3148 and I doubt we'll ever change that back.
A hand rolled `malloc` will have a much bigger chance of being subtly wrong.

codefromthecrypt: @aykevl Thanks. We we were missing the context

What you said "malloc and free has been safe for use since https://github.com/tinygo-org/tinygo/pull/3148 and I doubt we'll ever change that back." should be in the PR desc as rationale to use the CGO as the only way and remove the others, especially as the benchmarks don't show any perf regression. The same PR can update the site docs to only say one way (CGO malloc/free). @lburgazzoli can you update https://github.com/tetratelabs/wazero/pull/1390 accordingly? Once that's in, I'll we can remove the manual exports from tinymem, leaving it only used for slice helper functions.

aykevl: > What you said "malloc and free has been safe for use since [tinygo-org/tinygo#3148](https://github.com/tinygo-org/tinygo/pull/3148) and I doubt we'll ever change that back." should be in the PR desc as rationale to use the CGO as the only way and remove the others, especially as the benchmarks don't show any perf regression.

@codefromthecrypt  I'm not sure I follow. Was this a reply to me, or which PR are you referring to?

deadprogram: > @lburgazzoli can you update https://github.com/tetratelabs/wazero/pull/1390 accordingly?

I think @codefromthecrypt was asking @lburgazzoli to update tetratelabs/wazero#1390 accordingly. :smile_cat:

content/docs/guides/webassembly/resources.md
```diff
@@ -0,0 +1,29 @@
...
+https://www.fermyon.com/blog/tinygo-webassembly-favicon-server
+
+**WasmEdge Runtime - Go**
+https://wasmedge.org/book/en/write_wasm/go.html
+
+**WASI and Node.js**
+https://k33g.hashnode.dev/series/wasi-nodejs
+
+**Wazero, first steps**
+https://k33g.hashnode.dev/series/wazero-first-steps
```
aykevl, line 32: Some things I noticed:

```go
// readBufferFromMemory returns a buffer from WebAssembly
func readBufferFromMemory(bufferPosition *uint32, length uint32) []byte {
    subjectBuffer := make([]byte, length)
    pointer := uintptr(unsafe.Pointer(bufferPosition))
    for i := 0; i < int(length); i++ {
        s := *(*int32)(unsafe.Pointer(pointer + uintptr(i)))
        subjectBuffer[i] = byte(s)
    }
    return subjectBuffer
}
```

This is one example where `unsafe.Slice` would be perfect (although I'm not sure about the uint32 vs byte conversion, or what is supposed to be happening there):

```go
// readBufferFromMemory returns a buffer from WebAssembly
func readBufferFromMemory(bufferPosition *byte, wordLength uint32) []byte {
    length := wordLength * 4
    subjectBuffer := make([]byte, length)
    copy(subjectBuffer, unsafe.Slice(bufferPosition, length))
    return subjectBuffer
}
```

aykevl (commented):
I didn't fully look at the new blog post series.
I'm also not sure whether it's worth it to review each and everyone of them. Instead, I think it's more important to provide good primary documentation in the first place, which, to be very honest, we have very little of. So, I guess that's a weak LGTM on this PR from me.

content/docs/guides/webassembly/resources.md
deadprogram: Removed from this PR, can add to this page once it is updated, I suppose.

deadprogram:
Thanks everyone for the feedback. Now merging.

PR #353, opened 2023-04-28
docs/reference/devices: update list of supported devices

deadprogram:
This PR updates the list of supported devices in the TinyGo drivers repo.

Part of this change is making this list here the canonical list, removing it from the README in the drivers repo. That way there is only one place to maintain.


aykevl:
Good idea, thanks!

PR #358, opened 2023-05-18
all: Update docsy 0.6, correct sidebar link

deadprogram:
This PR updates the Docsy theme to the latest released version v0.6.0

It also corrects the sidebar links for "edit page" so that they point to the correct location. See https://github.com/google/docsy/issues/1420#issuecomment-1503810483

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,6 +1,7 @@
 baseURL = "https://tinygo.org/"
 title = "TinyGo"
-languageCode = "en-us"
+languageCode = "en"
+
 theme = "docsy"
 
 enableGitInfo = true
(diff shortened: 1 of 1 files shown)


aykevl (approved):
LGTM.

Does this have a practical benefit, besides staying up to date?

deadprogram:
A few things appear to be simplified, and it is using the latest fontawesome. Nothing really obvious other than if you download the latest hugo and use this latest docksy everything runs as you expect without extra warnings.

deadprogram:
Thanks for review @aykevl now merging.

aykevl:
> if you download the latest hugo and use this latest docksy everything runs as you expect without extra warnings.

Sounds like a good improvement then :)

PR #362, opened 2023-05-23
guides/embedded: add resources page

deadprogram:
This PR adds a new section for guides/embedded with a new resources page with links to some of the official resources for embedded development.

### DIFF
--- a/content/docs/guides/embedded/_index.md
+++ b/content/docs/guides/embedded/_index.md
@@ -1,0 +1,7 @@
+---
+title: "Embedded"
+weight: 2
+description: >
+  TinyGo can compile code that can run on even very small microcontrollers.
+---
+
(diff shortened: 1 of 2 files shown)


aykevl:
While these are good suggestions, I think they could be a bit more specific. For example, for using the internal flash of a device it would be much more useful to point to an actual example of how to do that.

Also, mostly unrelated to this PR: we don't have a section for the new embedded flash API in the machine package documentation. It would be great to have it included in the documentation.

deadprogram:
This branch requires that #363 be merged first, and then this branch rebased.

deadprogram:
> While these are good suggestions, I think they could be a bit more specific. For example, for using the internal flash of a device it would be much more useful to point to an actual example of how to do that.

This PR is a starting point for more docs on these subjects. Right now there is no mention of them anywhere on the site, so it is at least an improvement. A future PR or PRs can add more detail.

aykevl (approved):

deadprogram:
Merging to use as basis for some further changes in the new section. Thanks for review @aykevl

PR #365, opened 2023-06-11
imports: always run tests on linux/amd64

aykevl:
This makes test outputs more consistent, even though it is a lot slower on my arm64 laptop (running the tests in QEMU).


deadprogram:
Meant to merge this already, thanks @aykevl

PR #369, opened 2023-06-14
home: remove link-down

deadprogram:
This PR should remove the broken "link-down" on the home page.


deadprogram:
It works now! Merging.

PR #370, opened 2023-07-27
fix(wasi): MD formatting for Links

shantanoo-desai:
Use Markdown Syntax for Links in the WASI doc.


deadprogram:
Thank you for the edit @shantanoo-desai now merging.

PR #371, opened 2023-08-11
Document watchdog peripheral

kenbell:
See https://github.com/tinygo-org/tinygo/pull/3855

### DIFF
--- a/content/docs/reference/machine.md
+++ b/content/docs/reference/machine.md
@@ -320,6 +320,34 @@ func (uart *UART) WriteByte(c byte) error
 
 Write a single byte to the UART output.
 
+## Watchdog
+
+```go
+type WatchdogConfig struct {
+	TimeoutMillis uint32
+}
+```
+
+The `WatchdogConfig` struct contains configuration for the watchdog peripheral.
+
+  * `TimeoutMillis` is the requested maximum delay between calls to `Update`.
+
+```go
+func (wd *Watchdog) Configure(config WatchdogConfig) error
+```
+Configures the watchdog.
+
+This method should not be called after the watchdog is started and on some platforms attempting to reconfigure after starting the watchdog is explicitly forbidden / will not work.
+
+```go
+func (wd *Watchdog) Start() error
+```
+Starts the watchdog.  After calling this method, `Update` must be called periodically.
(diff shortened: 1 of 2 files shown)


deadprogram:
@kenbell now that the Watchdog was merged, can you please update these docs based on that final version?

kenbell:
@deadprogram - sure thing. done

deadprogram:
Thanks for updating @kenbell 

Note that I switched the PR to point to `dev` branch.

Now merging, thank you!

content/docs/reference/machine.md
```diff
@@ -320,6 +320,34 @@ func (uart *UART) WriteByte(c byte) error
...
+
+```go
+func (wd *Watchdog) Start() error
+```
+Starts the watchdog.  After calling this method, `Update` must be called periodically.
+
+```go
+func (wd *Watchdog) Update()
+```
+Updates the watchdog, indicating that the app is healthy.  This method must be called periodically to prevent the CPU from resetting.
```
aykevl, line 350: The only suggestion I have is to give some alternative names for what this is called. For example, it's also called feeding (for example, [here](https://docs.micropython.org/en/latest/library/machine.WDT.html)), petting, or restarting. Some also call it kicking but I don't like the idea of kicking dogs :astonished:  

(Not necessary, just something that might help people that already know watchdogs from other APIs).

content/docs/reference/machine.md
```diff
@@ -320,6 +320,34 @@ func (uart *UART) WriteByte(c byte) error
...
+
+```go
+type WatchdogConfig struct {
+	TimeoutMillis uint32
+}
+```
+
+The `WatchdogConfig` struct contains configuration for the watchdog peripheral.
+
+  * `TimeoutMillis` is the requested maximum delay between calls to `Update`.
```
aykevl, line 333: `Millis` is probably just fine. Nanoseconds are used in many places so might have been better for consistency but considering the usual range of watchdogs, milliseconds are probably just fine in practice.

aykevl (commented):
I had wanted to take a look at the API before it was merged, but just saying now that it looks good to me. Nice and simple.

PR #372, opened 2023-09-03
fix: installation repo links for arch

occamist:
fixes installation repo links for arch, I noticed this on my new machine while installing tinygo :+1:  @ManoloEvans


aykevl (approved):
Looks like this is a recent change. Source: https://lists.archlinux.org/archives/list/arch-dev-public@lists.archlinux.org/thread/SHCCCKHLIWPXPOIR5ARNJKSVWL2SSUO7/

> The [community] repository will be merged into [extra] and will therefore be empty after the migration.

PR #375, opened 2023-10-01
Fixes to LLVM related installation

aykevl:
* Use newer LLVM versions
* Use `brew install llvm@16` instead of `brew install llvm` to avoid confusion with newer LLVM version (like now, with LLVM 17).
* Fix Fedora install command (@QuLogic: FYI)


ysoldak (approved):
yes, I had to do `brew install llvm@16` on my mac. I confirm that's correct command.

deadprogram:
Thanks @aykevl for the fixes and to @ysoldak for the review. Now merging.

PR #376, opened 2023-10-20
Document machine.DeviceID

kenbell:
See https://github.com/tinygo-org/tinygo/pull/3968


deadprogram:
Thanks @kenbell now merging

PR #379, opened 2023-11-04
docs(macos): update links according to HCI guidelines for improved readability

coolaj86:
## Preview

See the changes in action at:

<https://github.com/coolaj86/tinygo-site/blob/patch-1/content/getting-started/install/macos.md>

## Here, here!

I changed the links by
- making the visual target much larger by making the text of the links more specific and clear
- placing links closer to the beginning or end of the sentence rather than in the middle
- using strict markdown link indicators `<>` 

As general (and highly effective) HCI / Technical Writing / Accessibility guidelines, the words "this", "here", "click", etc should be hunted down and vanquished! (and replaced with action phrases / content descriptions)

## The Colors Duke, The Colors!

I'm a superchromat (i.e. perfect color / hue distinction), on a Macbook (nearly perfect hue/color display), and the difference between the space black text and dark grey blue links is so difficult to distinguish that I couldn't recognize the links as such.

I'd highly recommend changing both the unvisited and visited link colors and adding underlines to at least links in text bodies.

### DIFF
--- a/content/getting-started/install/macos.md
+++ b/content/getting-started/install/macos.md
@@ -7,7 +7,7 @@ description: >
   macOS install guide
 ---
 
-This page has information on how to install and use TinyGo on macOS. If you wish to build TinyGo from source, for example if you intend to contribute to the project, please take a look [here](../../../docs/guides/build).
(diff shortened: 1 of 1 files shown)


deadprogram:
@coolaj86 I think this is an excellent set of changes. Any chance you could give the same treatment to the other `getting-started` pages? :smile_cat:

PR #382, opened 2023-12-21
[Game Boy Advance] Add examples section

scraly:
In the documentation there is a command to build  gba-display app/program for the GBA but what is gba-display? 

It's an example, so the goal is to give the links to gba-display and other examples


deadprogram:
Thanks for adding this @scraly now merging.

PR #386, opened 2024-01-16
Solving typo in build.md

DiegoAsterio:
The version for llvm apt repository did not correspond the llvm version one would  install in the subsequent command. 

```
sudo apt-get install clang-16 llvm-16-dev lld-16 libclang-16-dev
```

I was having issues following the guide. After this change I'm able to install `clang-16`


aykevl:
This has been solved on https://tinygo.org, so is no longer necessary.

PR #387, opened 2024-01-27
Adding rotary encoder to list of devices

bgould:
Adding rotary encoder to list of devices once https://github.com/tinygo-org/drivers/pull/648 is finalized.


aykevl:
Can you also update the count of supported devices at the top?

bgould:
Good catch! Updated :)

deadprogram:
Thanks @bgould now merging.

PR #389, opened 2024-02-25
Revisit build fix PR

soypat:
@aykevl


deadprogram:
Thanks for the fixes @soypat and @aykevl now merging.

PR #391, opened 2024-03-25
Update to Docsy 0.7

aykevl:
Most notably, this is an update to Bootstrap 5.2. It only needed a single line change to fix the layout.

It would make it slightly easier to add a TinyGo Playground on the homepage (just like on https://go.dev/) because then I don't have to juggle with Bootstrap versions.

I tried updating to Docsy 0.8, but the search function breaks with that change and I'm not sure how to fix it. So I decided to just bump the version to 0.7.

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -28,7 +28,7 @@ Ready to get started? [Click here](getting-started).
 
 {{% /blocks/lead %}}
 
-{{< blocks/section color="primary-light" >}}
+{{< blocks/section color="primary-light" type="row" >}}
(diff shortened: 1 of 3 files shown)


deadprogram:
There is a change to the font weight on the homepage, but that is not a big deal.

Merging, thanks @aykevl

aykevl:
> There is a change to the font weight on the homepage, but that is not a big deal.

Yeah, I think it's just a theme change (not a bug). There are many small changes, that one is probably the biggest one.

Thanks for merging :)

PR #395, opened 2024-03-29
Add a playground view on the homepage

aykevl:
This is the result of what I described previously on Mastodon: https://hachyderm.io/@ayke/112157503770449896

Screenshot_20240325_182034

Eventually I'd like to add the Gopher Badge as well so we can have an appropriate simulator, but I think the Arduino Uno and the BBC micro:bit are already nice to showcase in this way.

This will very likely put more strain on the Google Cloud Run app that compiles the source code. I'll have to monitor it to see whether it causes any issues.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "static/playground"]
+	path = static/playground
+	url = https://github.com/tinygo-org/playground.git
--- a/config.toml
+++ b/config.toml
@@ -91,3 +91,12 @@ enable = true
 # change content/<lang>/... to content/...
 from = "content/(.*?)/(.*?)"
 to = "content/$2"
+
+# Add some extra headers to the development server.
+# This is only relevant with `hugo serve`.
+[server]
+[[server.headers]]
+    for = '/**'
+    [server.headers.values]
+      Cross-Origin-Opener-Policy = "same-origin"
+      Cross-Origin-Embedder-Policy = "require-corp"
--- a/content/_index.md
+++ b/content/_index.md
@@ -28,7 +28,66 @@ Ready to get started? [Click here](getting-started).
 
 {{% /blocks/lead %}}
 
-{{< blocks/section color="primary-light" type="row" >}}
+{{% blocks/section color="primary-light" %}}
+<link rel="stylesheet" href="playground/simulator.css">
+<script type="module" src="playground.js"></script>
+<div class="col">
+	<div class="container" id="playground">
+		<h1 class="text-center">Try TinyGo</h1>
+		<div class="row px-0">
+			<div class="col col-auto">
+				<div class="input-group mb-3">
+					<span class="input-group-text">Example</span>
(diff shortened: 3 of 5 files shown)


deadprogram:
This is very cool!

The original Arduino is a old-school classic.
The BBC Microbit is also a classic. Perhaps the demo for the microbit should stick to the onboard LED matrix? I think very few people have ever hooked up a display to one.
Maybe it would be better to add another board, one with an actual onboard display?
Also, the Circuit Playground is a really popular board. It would be great to include that as well as a board with a display.
I do understand having the playground on the home page does not preclude keeping our more "fully stocked" playground as well.

Anyhow, just a few thoughts. Overall it is really neat.

aykevl:
> Perhaps the demo for the microbit should stick to the onboard LED matrix? I think very few people have ever hooked up a display to one.

Yeah good idea, though it will probably require some modifications to microbitdisplay and the playground for that display to work well. It's normally bitbanged which is difficult to simulate correctly.

> Maybe it would be better to add another board, one with an actual onboard display?

I'm working on adding the Gopher Badge :) it'll require a TinyGo change though.

> Also, the Circuit Playground is a really popular board.

Good idea, any suggestions what the example could be? Perhaps just animating the RGB LEDs?

> I do understand having the playground on the home page does not preclude keeping our more "fully stocked" playground as well.

Exactly. Though perhaps it would be best to actually integrate that playground into the website (like https://go.dev/play is) - right now it's an entirely standalone thing with different theming an all. I guess I can add https://tinygo.org/play in a later PR and redirect https://play.tinygo.org to it?

Some other ideas:

  * Add a "tour of TinyGo" similar to https://go.dev/tour/ (any volunteers for that?)
  * Integrate the playground into the board packages, so that we can for example show a simulator on https://tinygo.org/docs/reference/microcontrollers/circuitplay-express/.

aykevl:
I have updated the PR:

 * I've added two new examples: one with a Circuit Playground Express (showing the RGB LEDs) and one with the Gopher Badge (showing the display).
 * I've removed the BBC micro:bit example with the external display. Maybe we can add it later with proper display support.

I am not so sure about the Gopher Badge example. I can maybe just copy the example from play.tinygo.org, or maybe you have better suggestions? (Also tagging @conejoninja)

The Gopher Badge display is currently rather slow. A new API will greatly speed this up, see: https://github.com/tinygo-org/tinygo/pull/4224

deadprogram:
https://github.com/tinygo-org/tinygo/pull/4224 has been merged into the `dev` branch in that repo now.

aykevl:
> [tinygo-org/tinygo#4224](https://github.com/tinygo-org/tinygo/pull/4224) has been merged into the `dev` branch in that repo now.

Thanks, I've deployed the patch and verified that it works! (This PR doesn't need an update).

deadprogram:
This is totally awesome @aykevl thank you for working on it! Now merging.

PR #396, opened 2024-04-17
docs: add extism as runtime for tinygo wasm modules

nilslice:


deadprogram:
Thank you for the PR @nilslice I switched it to target the `dev` branch to help keep it orderly with a few other PRs that are also about to land.

Now merging, thanks again!

PR #397, opened 2024-04-17
Fix doc-gen and update docs

aykevl:
With a few simple tweaks I got doc-gen working again (though finding these issues took some time). The doc-gen script now works with Go 1.22.


deadprogram (approved):
At last! Thank you very much for fixing this @aykevl now merging.

PR #398, opened 2024-05-01
Update to Docsy v0.9.1

aykevl:
I'm doing this to stay up to date, and because I'm looking forward to dark mode support in Docsy v0.10 (which according to https://github.com/google/docsy/issues/331 seems to be almost finished).

There are a few small visual changes, which I think are generally an improvement:

  - Buttons are more square.
  - Links now have underscores, for accessibility reasons.
  - The ellipsis after "Read more..." on the home page is gone.

I did a quick browse through the website and couldn't find any breaking changes.

Note that not all hugo versions seem to work. My Fedora-shipped version v0.111.3 resulted in errors, and the latest version also results in errors for me. However v0.122.0 works for me.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -50,8 +50,6 @@ breadcrumb_disable = false
 sidebar_search_disable = true
 #  Set to false if you don't want to display a logo (/assets/icons/logo.svg) in the top nav bar
 navbar_logo = true
-# Set to true to disable the About link in the site footer
-footer_about_disable = false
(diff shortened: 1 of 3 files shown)


aykevl:
Huh, it also fails in Netlify. Probably because of the Hugo version requirement.

aykevl:
Let's do this some other day when these issues are hopefully fixed.

aykevl:
See #407 for version 0.10.0.

PR #399, opened 2024-05-08
content: updated switch emulator recommendation since yuzu is no longer

23RoMax:
Since yuzu has been shutdown: https://www.polygon.com/24090351/nintendo-2-4-million-yuzu-switch-emulator-settlement-lawsuit and a fork is available, I thought an update to the recommended testing tool might be due.


deadprogram (approved):
Thanks for the edit @23RoMax :crying_cat_face:

PR #400, opened 2024-05-10
Update wioterminal.md adding extra info to help people with ought automounting

Echo-Head-Wall:
Added extra info about what directory is needed for flashing so users with ought auto mounting know why the commands breaking and have a easier time finding out how to solve it.
i can also add in a example if needed though it would vary per distro so i left it out for now opting for a minimal note.


sago35 (approved):

deadprogram:
Thanks for the extra info @Echo-Head-Wall and to @sago35 for review. Now merging.

PR #401, opened 2024-05-24
reference: add `-serial=` flag documentation

aykevl:
I realized we don't actually have documentation for the `-serial=` flag. So here it is.


deadprogram:
Thanks for the additions @aykevl now merging.

PR #402, opened 2024-06-02
fix: wireless and storage link ref pointing to correct part of documentation in embedded-resources

23RoMax:
The reference to the wireless and the data storage section was pointing to the sensors sections. I took the liberty to fix it while reading through it.


deadprogram:
Thanks for the corrections @23RoMax now merging.

PR #403, opened 2024-06-06
Tour of TinyGo

aykevl:
Add a 'tour of TinyGo' with interactive examples in a mini-playground.

Work in progress.

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -48,6 +48,7 @@ Ready to get started? [Click here](getting-started).
 			</div>
 			<div class="col col-auto">
 				<button class="btn btn-secondary playground-btn-flash" disabled>Download binary</button>
+				<a href="/tour/" class="btn btn-link">Tour of TinyGo</a>
 			</div>
 		</div>
 		<textarea placeholder="Loading..." class="form-control input" rows="20" style="font-family: monospace; tab-size: 4" spellcheck="false"></textarea>
--- a/content/getting-started/tour.md
+++ b/content/getting-started/tour.md
@@ -1,0 +1,8 @@
+---
+title: "Tour"
+type: "docs"
+---
+
+We have a Tour of TinyGo, where you'll learn how to use TinyGo even if you aren't very familiar with programming hardware. You can find it [here]({{< ref "tour/blink/onboard.md" >}}).
+
+The tour assumes you are already familiar with Go. If not, you can follow the [Tour of Go](https://go.dev/tour/) to get started.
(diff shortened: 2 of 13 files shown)


aykevl:
Ready for review.

A few notes:

  - This tour should be extended with lots of things, it's just to get things started. For example: using a button as a switch, debouncing, using a display, controlling WS2812 LEDs, etc.
  - I've put a link on the homepage with the playground there, but I'm not sure about the design. Suggestions?

content/_index.md
```diff
@@ -48,6 +48,7 @@ Ready to get started? [Click here](getting-started).
 			</div>
 			<div class="col col-auto">
 				<button class="btn btn-secondary playground-btn-flash" disabled>Download binary</button>
+				<a href="/tour/">Tour of TinyGo</a>
```
deadprogram: I would instead add this at the bottom of this file something like this:

```markdown
{{% blocks/feature icon="fa fa-map-signs" title="Tour of TinyGo" url="/tour/" %}}
Take a tour of TinyGo online and try it out.
{{% /blocks/feature %}}
```

aykevl: Unfortunately that doesn't quite work: it looks like that section is only made for 3 features and with 4 it will wrap around:

Screenshot_20240616_173018

We could do this, but I think it looks weird.

However, I have improved the link slightly with a small change:

Screenshot_20240616_173101

aykevl: I was going for something like this (on the https://go.dev homepage) but in my opinion looking better.

Screenshot_20240616_173316

content/_index.md
```diff
@@ -48,6 +48,7 @@ Ready to get started? [Click here](getting-started).
 			</div>
 			<div class="col col-auto">
 				<button class="btn btn-secondary playground-btn-flash" disabled>Download binary</button>
+				<a href="/tour/" class="btn btn-link">Tour of TinyGo</a>
```
deadprogram, line 51:
```suggestion
                               {{< button href="/tour/" class="btn btn-secondary" >}}Tour of TinyGo{{< /button >}}
```

aykevl: Are you sure about that?

```
Error: error building site: process: readAndProcessContent: "/home/ayke/src/tinygo/site/content/_index.md:31:1": failed to extract shortcode: template for shortcode "button" not found
```

I prefer a link-style button: to me, buttons are something you do on a given page (like in a webapp for example), and links lead to a different page. This one clearly goes to a different page.

deadprogram: Let's just go with what you have now, and then we can change later if we thing we need to.

PR #404, opened 2024-06-09
docs: add tau as runtime for tinygo wasm modules

samyfodil:


deadprogram:
@samyfodil I think this might be better as an addition to https://github.com/tinygo-org/awesome-tinygo what do you think?

samyfodil:
@deadprogram Just like extism, tau runs wasm modules compiled with tinygo. it also uses tinygo as default wasm compiler for go. This is why I thought adding it here.
I don't think it really fits in [awesome-tinygo](https://github.com/tinygo-org/awesome-tinygo) because tau itself can't be compiled with tinygo.

deadprogram:
In that case, please keep the list in alpha order.

samyfodil:
@deadprogram done :+1:

deadprogram:
Thanks for the update @samyfodil now squash/merging.

PR #407, opened 2024-06-12
Update to Docsy v0.10.0

aykevl:
This adds support for light/dark mode, which I've waited for for a while. (I wish we could just make it automatic without menu entry, but this is already better than nothing).

There are a few small visual changes, which I think are generally an improvement:

  - Buttons are more square.
  - Links now have underscores, for accessibility reasons.
  - The ellipsis after "Read more..." on the home page is gone.
  - The copyright in the footer has changed slightly.

I did a quick browse through the website and couldn't find any breaking changes.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -50,10 +50,10 @@ breadcrumb_disable = false
 sidebar_search_disable = true
 #  Set to false if you don't want to display a logo (/assets/icons/logo.svg) in the top nav bar
 navbar_logo = true
-# Set to true to disable the About link in the site footer
-footer_about_disable = false
 # Set the maximum number of menus to be displayed in the sidebar
 sidebar_menu_truncate = 1000
+# Show a menu to the left of the search menu to toggle light/dark mode (or set it to auto).
+showLightDarkModeMenu = true
 
 [params.links]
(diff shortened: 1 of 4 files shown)


aykevl:
And now it actually builds, after a few changes.
It looks like it uses a rather old build environment, @deadprogram do you know anything about that? I had to pick a different Go version because by default it would use Go 1.12.

deadprogram:
We need to update the `netlify.toml` file:

I think this is what we need to have in there:

```toml
[build.environment]
  HUGO_VERSION = "0.127.0"
  GO_VERSION = "1.21"
  NODE_VERSION = "v20.x"
```

aykevl:
@deadprogram I mean the build image: https://docs.netlify.com/configure-builds/overview/#build-image-selection
Apparently this can be configured from the web UI somewhere.

deadprogram:
I think we are already using the latest image:

image

aykevl:
Oh, okay. I think I saw Go 1.12 in there, which is kinda old, but 20.04 ships Go 1.13 so it's not that far off.
Then I think there's nothing we need to update.

(That means this PR is ready for merge in my opinion).

deadprogram:
OK, now merging. Thanks for the update @aykevl

PR #411, opened 2024-07-02
Fix light/dark mode syntax highlighting

aykevl:
See: https://www.docsy.dev/docs/adding-content/lookandfeel/#lightdark-color-themes

Previously the code would effectively be in dark mode, no matter whether the website was light or dark mode. This led to situations like here, which I don't think looks very good:

Screenshot_20240702_183242

After this change, the syntax highlighting theme changes with the website theme:

Screenshot_20240702_183442
Screenshot_20240702_183458

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -1,0 +1,1 @@
+@import 'td/code-dark'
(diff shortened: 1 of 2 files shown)


deadprogram:
Thanks for the fix @aykevl now merging.

PR #412, opened 2024-07-06
Syntax highlighting on the playground

aykevl:
This adds syntax highlighting to the playground on the homepage and in the tour.

I picked highlighting themes that match the code samples on the website as much as possible, although it's not a perfect match. CodeMirror includes a oneDark theme (same as on the website) for dark mode so that's great, but for light mode I actually created a new CodeMirror theme for the tango theme that is used by default by Docsy.

I have tried to keep the appearance and accessibility properties of the textarea the same as much as possible. I think the outcome is pretty neat: it looks just like any other Bootstrap input but with numbered lines and syntax highlighting.

~Eventually I'd like to add a feature to show compiler error messages where they happen. Or at least to highlight the lines with errors. But that's for another time.~ implemented

One major reason for choosing CodeMirror 6 is that it works well on mobile, while most of these web code editors don't.

I've also fixed an issue I introduced in #411: the tab size is now back to 4 spaces instead of 8.

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -1,1 +1,5 @@
-@import 'td/code-dark'
+@import 'td/code-dark';
+
+code {
+    tab-size: 4;
+}
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -3,3 +3,38 @@
 code {
     tab-size: 4;
 }
+
+.playground-editor {
+    /* make the editor scrollable */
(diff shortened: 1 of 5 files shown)


aykevl:
Updated. It now also shows errors where they happen, almost like an IDE. I have plans to make these errors more accurate, to bring truly IDE-like error reporting to the editor.

deadprogram:
Works exactly as expected! Thank you very much for working on this @aykevl now merging.

PR #416, opened 2024-07-22
playground: style simulator like a Bootstrap card

aykevl:
In my opinion, this looks _much_ better than the previous VSCode-like style.
  * It supports better keyboard navigation of the UI (not complete, but better than it was before).
  * It now supports dark mode (light+dark), instead of staying dark all the time.
  * It integrates with the look&feel of the rest of the website.

This is what it now looks like, which I think is a lot cleaner:

Screenshot_20240722_192254


deadprogram:
For sure looks a lot better, thank you @aykevl now merging.

PR #417, opened 2024-07-25
Simplify use of docsy module

deining:
This PR remove module `github.com/google/docsy/dependencies` which is not needed any more.
It also brings `README.md` up to date.
It also bumps various dependencies to their latest versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 public
 ./build
 node_modules
+resources
 npm-debug.log
+.hugo_build.lock
--- a/README.md
+++ b/README.md
@@ -28,11 +28,7 @@ Change directories into the tinygo-site directory:
 
     cd tinygo-site
 
-Install the "Docsy" theme:
-
-    git submodule update --init --recursive
-
-Now you should be able to run the site locally:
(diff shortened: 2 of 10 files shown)


deadprogram:
@deining thank you very much for this update. I notice that there is some issue with an unpopulated iframe at the very bottom of the page in the preview site at https://deploy-preview-417--tinygo.netlify.app/

What do you think?

deining:
> @deining thank you very much for this update.

You are welcome!

> I notice that there is some issue with an unpopulated iframe at the very bottom of the page in the preview site at https://deploy-preview-417--tinygo.netlify.app/
> What do you think?

Don't be deceived here, this iframe is an artefact which is present only in Netlify preview, you won't see in the deployed site.
If you scroll inside this iframe, you get an indication that this artefact is introduced from Netlify:

grafik

deining:
Have a look [here](https://deploy-preview-2071--docsydocs.netlify.app/):

grafik

That's how this iframe is supposed to look like. No idea why it doesn't show up like that in the preview of the tinygo site.

deadprogram:
Thanks for the clarification @deining now merging.

PR #418, opened 2024-07-25
Fix typos

deining:
This PR fixes various typos I spotted in the project.


deadprogram:
@deining the general idea of fixing typos is good, but there some things you need to know about. One is that a lot of the machine package docs are generated from the info in the main repo. See https://github.com/tinygo-org/tinygo-site/tree/release/doc-gen

As a result, many of your proposed changes would be overwritten when the docs are auto-updated, which will happen as part of this new release.

aykevl:
Merged, since this still seems like a good idea. A bunch of these will be overwritten, but that doesn't seem too bad to me. 

@deining if you are interested, the machine/*.md files are based on files in the tinygo repository. If you want to, you can fix the typos there (in src/machine).

PR #419, opened 2024-08-14
drivers: add redirect for all missing drivers

deadprogram:
This PR adds the custom domain redirect files for all drivers that were missing them. It also adds a missing `</html>` tag so they are now correctly formed.


aykevl (approved):
LGTM

deadprogram:
I do not think we need this, so closing.

PR #426, opened 2024-08-27
guides/webassembly: use `//export` instead of `//go:export`

aykevl:
`//export` is the official 'big Go' pragma, `//go:export` only exists as an alias because I was lazy in the past and now it's probably too late to remove that pragma. But we should at least use the correct one in the documentation.


deadprogram:
Thanks for that correction @aykevl I had it reversed in my brain. Now merging.

PR #431, opened 2024-09-21
Update on microcontrollers/arduino-mega2560, I2C support corrected

habdig7oficial:
Hello after, a discussion in the tinygo slack, one of the maintainers said that the I2C support on [arduino-mega2560](https://tinygo.org/docs/reference/microcontrollers/arduino-mega2560/) was wrong indicating that it have support when it doesn't so I corrected the documentation page

image

image

I look forward to I2C support on tinygo, but until there the documentation page needs to be corrected, later i will see if I can use it with arduino-uno whose page also says to have I2C.

Thank you all for the support! :)

### DIFF
--- a/content/docs/reference/microcontrollers/arduino-mega2560.md
+++ b/content/docs/reference/microcontrollers/arduino-mega2560.md
@@ -14,7 +14,7 @@ Note: the AVR backend of LLVM is still experimental so you may encounter bugs.
 | GPIO      | YES | YES |
 | UART      | YES | YES |
 | SPI       | YES | YES |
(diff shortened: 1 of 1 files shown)


aykevl:
Unfortunately this part of the documentation is autogenerated, so it will be overwritten on the next update.
It's still a bug, but ideally it should be fixed at the source: https://github.com/tinygo-org/tinygo-site/blob/release/doc-gen/main.go
(The issue is that it just checks whether the `I2C` type exists, it doesn't check whether there are actually any peripherals of the type `I2C`).

PR #434, opened 2024-10-23
Describe setup for Pico on Linux

tdunning:
I have added documentation about flashing the Pico on a Linux machine
with detailed commands and procedures. All commands and error messages
have been verified on my own machine.

### DIFF
--- a/content/docs/guides/tinygo-flash-errors.md
+++ b/content/docs/guides/tinygo-flash-errors.md
@@ -9,6 +9,58 @@ Flashing to a port may be a source of pain with a unconfigured computer. Here ar
 
 * [Serial port permission problems (Linux)](#permission-denied-linux)
 
+### Unable to locate any volume (Linux)
+
+This error happens with devices such as the Pi Pico (and the Pico-W) which use the UF2 protocol for flashing programs.  With this protocol, the device 
+emulates a USB block storage device that can be mounted. When you copy a properly formatted flash image onto this virtual disk drive, the side effect
+is to flash that image onto the device.
+
+For the pico with `target=pico`, `tinygo` expects to see this device in the current directory under the name `RPI-RP2`. If this volume is is not found,
+you will get an error message like this:
+```
+failed to flash /tmp/tinygo173718394/main.uf2: unable to locate any volume: [RPI-RP2]
+```
+
+To fix this problem, you need to mount the volume after restarting the Pico while holding down the `BOOTSEL` button. You can do this manually, but it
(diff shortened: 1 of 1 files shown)


content/docs/guides/tinygo-flash-errors.md
```diff
@@ -9,6 +9,58 @@ Flashing to a port may be a source of pain with a unconfigured computer. Here ar
...
+
+For the pico with `target=pico`, `tinygo` expects to see this device in the current directory under the name `RPI-RP2`. If this volume is is not found,
+you will get an error message like this:
+```
+failed to flash /tmp/tinygo173718394/main.uf2: unable to locate any volume: [RPI-RP2]
+```
+
+To fix this problem, you need to mount the volume after restarting the Pico while holding down the `BOOTSEL` button. You can do this manually, but it
+is better to edit the `/etc/fstab` file to allow users to do this themselves. You can do this by adding a line like this to `/etc/fstab`:
+```
```
soypat: add `bash` code identfier here

content/docs/guides/tinygo-flash-errors.md
```diff
@@ -9,6 +9,58 @@ Flashing to a port may be a source of pain with a unconfigured computer. Here ar
...
+LABEL=RPI-RP2  /path/to/volume/RPI-RP2  vfat	defaults,umask=000,users
+```
+where you replace `/path/to/volume` with a pathname of your choosing. It is common for this to be put in your home directory or in a system standard
+place such as `/mnt/RPI-RP2`, but it can also be in the directory where you expect to run `tinygo`. You will also need to create a directory with the path 
+`/path/to/volume/RPI-RP2`. Once you have done this and after restarting the Pico in
+`BOOTSEL` mode, you should be able to see the Pico's block device using this command
+```
```
soypat: `bash` identifier hre too and wherever you run commands to get code colorization

content/docs/guides/tinygo-flash-errors.md
```diff
@@ -9,6 +9,58 @@ Flashing to a port may be a source of pain with a unconfigured computer. Here ar
 
 * [Serial port permission problems (Linux)](#permission-denied-linux)
```
soypat, line 11: add a entry in the table here for the section you've added

tdunning:
Thanks for the suggestions. I will update the pull request soon.

tdunning:
Yay for weekends. 

Here are the updates to match the comments.

soypat (approved):
LGTM! Thanks for contribution

tdunning:
There will be more when I get further into my project!

PR #440, opened 2025-01-25
Be specific about universal location of wasm_exec.js

jphastings:
I spent a long while trying to figure out how to find `wasm_exec.js` for my build script. The `docker` command below this edited line implies a universal location (of `/usr/local/tinygo/targets/wasm_exec.js`) which isn't true on many systems (eg. installed from zip, or by Homebrew, or on Windows).

This change ensures the intended "correct" way to find the file is easily visible in the docs.


deadprogram (approved):
Thanks for that clarification @jphastings

PR #441, opened 2025-01-28
build: add lld dependency for Homebrew

aykevl:
This used to be part of the llvm package but is now separate, and needs to be installed separately.


deadprogram:
Thanks for the addition @aykevl now merging.

PR #442, opened 2025-02-10
media: adding a few newer articles and videos

deadprogram:
This PR adds a few of the more recent articles and videos. There are quite a few, this is just a selection of some that I could remember offhand.


conejoninja:
Thanks! merged.

PR #444, opened 2025-02-14
Make Qwiic port usage clear

jphastings:
It was unclear to me that neither `machine.I2C0` nor `machine.I2C1` work as-is with this device, so I've made a note clarifying that you _must_ configure `I2C1` to be able to use the Qwiic/StemmaQT port.


deadprogram:
Thanks for the improvement @jphastings now merging.

PR #445, opened 2025-02-22
content: add links to PIO support for all current rp2040 based boards

deadprogram:
This PR adds links to the PIO support for all current rp2040 based boards.

Fixes https://github.com/tinygo-org/tinygo/issues/2589


deadprogram:
Merging since is pretty non-controversial.

PR #447, opened 2025-03-05
Update playground

aykevl:
There are a bunch of fixes. One big one is the addition of some (privacy-preserving) tracking when the playground compiles a new binary to run in the browser. But there are also a bunch of other small fixes that require updating the HTML a little bit.

See: https://play.tinygo.org/stats/ (this page is very bare-bones, I'll update it in the future with more details once there's enough data to show anything).


deadprogram:
This is great work @aykevl thank you! Now merging.

PR #448, opened 2025-03-13
tour: add short WS2812 tour

aykevl:
I will likely add more at some point, going into detail on power consumption and other things like RGBW LEDs. But this should already be quite useful for people starting out with WS2812 LEDs.

Here is the Netlify preview: https://deploy-preview-448--tinygo.netlify.app/tour/ws2812/intro/


deadprogram:
This is really good info here @aykevl thank you for working on it! Now merging.

PR #451, opened 2025-04-07
Add driver design page

soypat:
recently was inspired to start working on protips for people designing drivers.

Still a WIP.

Would love a review and suggestions from driver developer extraordinaires @aykevl @deadprogram  @ysoldak  @sago35


amken3d:
Nice. This would hav helped me when I was writing the TMC drivers. I like the practical aspect of the write up. You should do something similar for tips and tricks for more tinygo topics.

ysoldak:
I like "To abstract `machine.Pin` one can define `type PinOutput func(level bool)` and `type PinInput func() bool` functional interfaces."
Another alternative is for us to have `drivers.Pin`, but that's a separate discussion and I don't think we come to any decision yet? Also, do we expect pins be pre-configured or a driver's Config method to set them up?
Anyway, the core advise is to avoid `machine` package, that's good 👍 
We can always update docs later when we decide on pins.

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,223 @@
...
+
+In go we are not aware of this since we never interact with the low level representation of a slice:
+```go
+type slice struct {
+    data      *type // Points to the first element of the slice:  &s[0]
+    length    int   // This is returned when calling builtin len: len(s)
+    capacity  int   // This is returned when calling builtin cap: cap(s)
+}
+```
+Why is it important to know slices are referential? Well it is because the compile is aware of this and will check
```
ysoldak:
```suggestion
Why is it important to know slices are referential? Well it is because the compiler is aware of this and will check
```

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,223 @@
...
+
+```go
+type Device struct {
+    spi      drivers.SPI
+    bufRead  [2]byte
+    bufWrite [2]byte
+}
+
+func (d *Device) readRegister(addr uint8) (value uint8, err error) {
+    d.bufWrite = [2]byte{addr, 0}
```
ysoldak: Does not this create a new slice? I mean, every time. Or is compiler smart enough to replace values in already allocated fixed size slice?

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,223 @@
...
+func (d *Device) Configure() error {
+    // Since the argument buffer is not passed into Tx call,
+    // data does NOT escape and may be stack allocated.
+    data := []byte{1, 2, 3, 4}
+    return d.writeData(CONFIG_REG, data)
+}
+
+func (d *Device) writeData(addr uint8, data []byte) error {
+    d.data[0] = addr
+    n := copy(d.data[1:], data)
```
ysoldak: swap 119 and 120?

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,223 @@
...
+One can write a driver that shares the underlying shared logic like so:
+
+```go
+package hsc
+
+// device implements the shared logic between Honeywell HSC pressure sensor varieties.
+type device struct {
+    // calibration factors.
+    k1, k2, k3 uint32
+    lastRead   uint32
```
ysoldak, line 181: A piece of advice (to add above, in the beginning of the guide) to separate "update" of value from "reading" value, client being responsible to poll Update regularly?

ysoldak:
Add links to these?
- https://tinygo.org/docs/concepts/compiler-internals/heap-allocation/
- https://go.dev/wiki/SliceTricks#additional-tricks

content/docs/guides/driver-design.md
soypat: I've added a comment to clarify this :)

soypat:
@ysoldak I've applied your suggestions- thank you so much!

content/docs/guides/driver-design.md
```diff
@@ -95,10 +99,13 @@ func (d *Device) writeRegister(addr, newValue uint8) error {
     return d.spi.Tx(d.bufWrite[:], d.bufRead[:])
 }
 ```
+| Note on Go's array type in TinyGo |
+|------|
+| Note the use of static arrays types like `[2]byte`. "Arrays", as they are called in Go, if small enough will always be stack allocated. TinyGo currently has a **Max Stack Object Size** configuration which runs around the 256 byte size. That said, this fact is not relevant in this case since the array is attached to the `Device` struct which almost certainly is heap allocated. Most drivers, if not all, initialize and use a pointer type (`*Device`). If the `Device` is heap allocated then the arrays it contains will also be heap allocated- and this is OK! Remember, we are trying to constrain our allocations, not completely eliminate heap allocations. We win if during the lifetime of the program there are **ZERO** heap allocations. Allowing heap allocations at initialization is a good practice and actually encouraged in NASA's coding standards for extraplanetary missions. |
```
ysoldak: Link to NASA coding standards? Can be educational!

soypat: Will add it
https://en.m.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,361 @@
...
+
+- One package per peripheral
+    - So if your peripheral is the MPU6050 accelerometer create a new package called `mpu6050`
+    - Sometimes the logic can be shared between a family of peripherals such as the AS5601 and AS5600. In these cases one usually creates a package that contains this shared logic and x's the shared character position, so `as560x`
+
+- Define a base type `Device` for the handle to the peripheral. All interaction with the peripheral will be via methods on this type.
+    - Store HAL required for peripheral operation here such as communication buses (I2C, SPI drivers) and pins
+    - **Avoid**: using TinyGo `machine` package constructs in your driver (read as *don't import `"machine"`*). Users of the Go language (not only TinyGo) also consume the drivers package and they can't compile a program that uses the `"machine"` package. To abstract `machine.Pin` one can define `type PinOutput func(level bool)` and `type PinInput func() bool` functional interfaces.
+
+- Define a `Config` struct for peripherals with many configuration options. See example of a Config struct in action: https://github.com/soypat/lora/blob/main/lora.go
```
conejoninja: I would point to a driver under the tinygo-org repo

content/docs/guides/driver-design.md
```diff
@@ -0,0 +1,361 @@
...
+    - Define a `DefaultConfig` at the package level so that the user may easily instantiate a configuration. Maybe add a parameter or two to `DefaultConfig` to allow the user to easily tweak the most important parameters that they may want to change.
+    - Are some configuration parameters codependent and require calculating? Add a method on the `Config` type to set them consistently! Adding a `Validation() error` method is also useful during the `func (d *Device) Configure(cfg Config) error` call as well as letting users read the validation logic and understand what is a correct configuration. It's a great source for intuition on how the driver works!
+- **Avoid**: Floating point operations. 
+    - Peripherals rarely work with or return floating point representations of their measurements. More often than not you'll find they work with integers. There's a reason for this: integer operation hardware is widespread. Most small processors do not support floating point operations, and even if they do it is much slower than modern desktop PCs compared with integer operations
```
conejoninja: Maybe mention that there are some *unofficial* interfaces for some types of sensors, ie. temp sensors usually returns milliC instead of C ¿?

soypat: I've added a comment noting that is still a WIP and where they can find the details, thanks for the tip

content/docs/guides/driver-design.md
soypat: yes! I've deleted that link and added a list of compliant drivers below

soypat: That said, sadly all the drivers are mine :(. Be great to start working on making other existing drivers compliant so we can expand the list

ysoldak (approved):
Ship it!

content/docs/guides/driver-design.md
deadprogram: I'm not sure we have resolved this question yet, no?

soypat: Which question?

soypat: I couldn't find other examples which do a good job to sticking to the guidlines except for the ones added in https://github.com/tinygo-org/tinygo-site/pull/451/commits/7905de79ebee91e2bd836f8e39d49747cfa675d1

soypat:
Found a couple typos and in the heat of it added a concurrency section which I felt was missing based on my experiences programming and using async drivers in TinyGo.

soypat:
Last call for comments before merging! Will wait until Monday :)

ysoldak (approved):

soypat:
Mergerino the pull requesterino

soypat:
Thank you @ysoldak, @conejoninja for the feedback <3

PR #453, opened 2025-04-23
Warning on building tinygo within go workspaces.

ofauchon:
Warning on building tinygo within go workspaces.


deadprogram:
Thanks for the docs @ofauchon hopefully that will help out others who run into this! Now merging.

PR #454, opened 2025-05-01
Add tour to explain PWM

aykevl:
This tour explains how to do the following:

  * Blink an LED (on supported hardware, this is mostly just SAMD21)
  * Fade an LED
  * Control servos (directly, and using the servo driver).

### DIFF
--- a/content/tour/pwm/blink.md
+++ b/content/tour/pwm/blink.md
@@ -44,7 +44,7 @@ Next up, we connect the PWM peripheral to the given output pin:
 
 We get a channel number back. One PWM peripheral typically has 2-4 channels, which can be connected to certain pins. Which pins can be used depends a lot on the hardware, which we'll [get to later](../multiple/#finding-the-correct-pwm-peripheral-and-pins). For now, all you need to know is that the PWM channel is now connected to a single GPIO pin.
 
-To actually make the LED blink, we will set the duty cycle to half the "top" value of the LED:
+To actually make the LED blink, we will need to set the duty cycle. The duty cycle refers to the part of the time when the PWM is turned on. In this case, we will set the duty cycle to half the "top" value of the LED:
 
 ```go
 	// Blink the LED, setting it to "on" half the time.
--- a/content/tour/pwm/blink.md
+++ b/content/tour/pwm/blink.md
@@ -1,0 +1,112 @@
+---
+title: "LED: blink"
+description: "Get a LED to blink using the PWM peripheral."
+weight: 1
+---
+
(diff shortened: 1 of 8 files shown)


content/tour/pwm/blink.md
```diff
@@ -0,0 +1,112 @@
...
+	ch, err := pwm.Channel(led)
+	if err != nil {
+		println("could not obtain channel:", err.Error())
+		return
+	}
+```
+
+We get a channel number back. One PWM peripheral typically has 2-4 channels, which can be connected to certain pins. Which pins can be used depends a lot on the hardware, which we'll [get to later](../multiple/#finding-the-correct-pwm-peripheral-and-pins). For now, all you need to know is that the PWM channel is now connected to a single GPIO pin.
+
+To actually make the LED blink, we will set the duty cycle to half the "top" value of the LED:
```
deadprogram: Perhaps define "duty cycle" here.

content/tour/pwm/servo.md
```diff
@@ -0,0 +1,183 @@
+---
+title: "Servo: intro"
+description: "Control a servo motor directly using PWM"
+weight: 10
+---
+
+Servos are controlled using a form of pulse-width modification (PWM) to control their position. Every 20ms or so (the precise interval doesn't matter too much) a pulse of 1ms to 2ms is sent that indicates the target position for the motor. The pulse width is what matters here: 1.5ms indicates the center position, 1ms all the way to the right, and 2ms all the way to the left.
```
deadprogram: Tell us what a servo is, please.

content/tour/pwm/blink.md
deadprogram:
```suggestion
To actually make the LED blink, we will need to set the duty cycle. The duty cycle refers to the part of the time when the PWM is turned on. In this case, we will set the duty cycle to half the "top" value of the LED:
```

content/tour/pwm/servo.md
deadprogram:
```suggestion
Servos are rotary motors that allow for precise control of their angular position. The position is set using a form of pulse-width modification (PWM). Every 20ms or so (the precise interval doesn't matter too much) a pulse of 1ms to 2ms is sent that indicates the target position for the motor. The pulse width is what matters here: 1.5ms indicates the center position, 1ms all the way to the right, and 2ms all the way to the left.
```

deadprogram:
@aykevl I have now made some specific suggestions on edits for this PR. PTAL!

deadprogram:
Now merging, thanks for working on this @aykevl

PR #456, opened 2025-06-26
Fix link to lsm6dsox

suhlig:


deadprogram:
Thanks for the correction @suhlig now merging.

PR #463, opened 2025-09-08
tour: add I2C tour component explaining the LIS3DH

aykevl:
This also updates the playground, to add I2C and specifically LIS3DH support.


deadprogram:
I wonder if `I²C` although correct, might be better just as the more common `I2C`?

deadprogram:
Also, it would be good to explain "What Is I2C" someplace in here.

deadprogram:
@aykevl please see my comments on this PR.

aykevl:
Updated the PR to use I2C instead.

aykevl:
@deadprogram ping ;)

deadprogram:
Looking very good thanks for making those changes @aykevl now merging

PR #464, opened 2025-09-23
microcontrollers: add I2C to listed pins

aykevl:
This makes it easier to find which pins support I2C and which peripheral is used. It uses the comments added in https://github.com/tinygo-org/tinygo/pull/4952 and https://github.com/tinygo-org/tinygo/pull/5053.

My plan is to refer to these pins in the documentation from the tour.


deadprogram:
Switched to `dev` branch and now merging. Thanks @aykevl

Issue #461, opened 2025-08-20
threading support controls? seems docs need update


glycerine:
https://github.com/tinygo-org/tinygo-site/blob/9d0a8d583a56e96307555221227c888baa2a783f/content/docs/guides/tips-n-tricks.md?plain=1#L11

says

"TinyGo code runs on a single core, in a single thread (think `GOMAXPROCS=1`).
Since scheduling in TinyGo is [cooperative]"

However, the latest release notes, under runtime, says

"darwin: add threading support and use it by default"

which seems to contradict the single threaded cooperative concurrency claim of the doc/web site.

Also it would be great to link to what settings are available/ how to control the new concurrency options.

aykevl:
You are right, see: https://github.com/tinygo-org/tinygo-site/issues/461

PR #465, opened 2025-09-29
tips-n-tricks: note that TinyGo can be multicore

aykevl:
See: https://github.com/tinygo-org/tinygo-site/issues/461


deadprogram:
Thanks for the addition. Now merging.

PR #466, opened 2025-09-30
Bump Docsy theme to latest released version v0.12.0

deining:
This PR bumps Docsy theme to latest released version v0.12.0.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,4 @@ module github.com/tinygo-org/tinygo-site
 
 go 1.22
 
-require github.com/google/docsy v0.11.0 // indirect
+require github.com/google/docsy v0.12.0 // indirect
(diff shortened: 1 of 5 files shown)


aykevl:
Looks like it isn't quite working yet. 

Before merging, did you check that the website is still working as expected? That means going though a bunch of pages to see whether the layout etc didn't change in an unexpected way. (Especially the playground and tour are important to check). Usually there are small changes that need to be made when bumping the version.

Also, are there features in v0.12 that we need?

deining:
> Looks like it isn't quite working yet.

You are right.

> Before merging, did you check that the website is still working as expected? That means going though a bunch of pages to see whether the layout etc didn't change in an unexpected way. (Especially the playground and tour are important to check).

I overlooked these regressions, sorry. I will now look into this.

> Also, are there features in v0.12 that we need?

I started the theme version bump because current head of repo does not build with latest released hugo version any more.

On another route: deployment of this PR on Netlify fails because of an outdated build image in Netlify's settings. Can you do me a favor and change the build image to latest released version Ubuntu 24.04? Thanks in advance!

PR #467, opened 2025-10-06
install/windows.md: document how to do persistent PATH changes

dkegel-fastly:
Note: the site is a bit out of date with respect to hugo, so to test this, I had to install an old version of hugo with
```
go install github.com/gohugoio/hugo@v0.140.2
```


aykevl (approved):
Yeah unfortunately hugo and Docsy tend to have backwards incompatible changes so the two versions need to stay somewhat in sync. But the PR looks good to me.

PR #469, opened 2026-01-02
content: mention tinygo-bin AUR package

jakebailey:
The main Arch package is still at 0.37.0; the AUR package that just uses the deb is at 0.40.1. I was recently made a co-maintainer of the package and am keeping it updated. I think it's worth mentioning.


deadprogram:
Thanks @jakebailey this is very helpful! Now merging.

PR #471, opened 2026-01-13
Added new articles and videos

conejoninja:
I could not test this change (but it's trivial enough) because the theme docsy is outdated and I couldn't downgrade my version of hugo.

There's a PR to update docsy ( #466 ), but we should check the deployment version of hugo too.

**Note**: It's been a while, should I merge against `release` or `dev` ?


deadprogram:
> Note: It's been a while, should I merge against release or dev ?

I usually merge to `dev` and then manually merge to release when ready.

conejoninja:
Changed the base branch

deadprogram:

PR #472, opened 2026-01-21
Added documentation for Vicharak Shrike Lite

YajTPG:
Related PR: https://github.com/tinygo-org/tinygo/pull/5170

Thanks!


deadprogram:
Thanks for the docs update @YajTPG 

Did you generate it using https://github.com/tinygo-org/tinygo-site/tree/release/doc-gen from the branch with your addition of the board itself?

If not, then this would be done once that PR lands. The other changes you added would then have a place they would not be overwritten when next regenerated.

Hope that makes sense!

YajTPG:
Ah, that's what I was missing the entire time!
It did occur to me it was a bit weird editing the machine doc manually but didn't think much of it, my apologies.

Generated the machine package using doc-gen in yajtpg@7e4d649
Also realized I was using the wrong target in https://github.com/tinygo-org/tinygo-site/pull/472/files#diff-d67896908291c57191890725855f28bd7293ff78958b094921bf2d06879fb40eR71, fixed that too.

Thank you for pointing that out, @deadprogram!

deadprogram:
Thanks for the changes @YajTPG now squash/merging.

PR #476, opened 2026-03-05
Updating debugging guide steps for Raspberry Pi Debug Probe

JamesBelchamber:
We spent a little time getting this working on Fedora and macOS, and landed up with some internal docs we figured we could just push upstream. Appendix A in "Getting Started with Raspberry Pi Pico" guide seems to be out of date and no longer works, but we have successfully used these instructions on macOS Tahoe, Fedora 43 and Ubuntu 24.04.

We also added a few notes on debugging since we found `-ocd-commands "adapter speed 5000"` needed to be added to `tinygo gdb` to get it working with the Debug Probe, and we used the other flags to debug this :)

### DIFF
--- a/content/docs/guides/debugging.md
+++ b/content/docs/guides/debugging.md
@@ -54,6 +54,31 @@ You can install the most common dependencies through Homebrew:
     brew tap ARMmbed/homebrew-formulae
     brew install arm-none-eabi-gcc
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you very much for the documentation improvements. I switched the target branch to `dev`. Now merging!

PR #477, opened 2026-03-08
docs: Add Helix IDE Integration instructions

shoenig:
Adding an IDE-Integrations page with instructions on getting TinyGo projects to work with the Helix editor. The Helix docs are a bit dense in this area (took me over an hour of fumbling around to figure this out), so it seemed like having an example could be helpful for others.


deadprogram:
I switched the branch for this PR to `dev` @shoenig 

Thanks for the addition! Now merging.

PR #478, opened 2026-03-08
docs: clarify hugo install instructions

shoenig:
Recent versions of Hugo throw an error when trying to build the `tinygo-site` repository. It seems `v0.139.5` was the most recent working version, so I've added a note in the README.md making note of this, and giving a 1-line `go install` command to use Go to install the correct version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,14 @@ https://github.com/google/docsy
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for the updated info @shoenig now merging.

PR #481, opened 2026-04-16
Update debugging.md to mention automake

tdunning:


deadprogram:
Thanks for the update @tdunning now merging.

PR #485, opened 2026-04-17
imports: replace deprecated ioutil, add 3-minute timeout per test

deadprogram:
This PR is to make some improvements to the `import` program that is used to update the page that show the latest `stdlib` build/test status.

- Replace ioutil.TempDir/WriteFile with os.MkdirTemp/os.WriteFile
- Use buf.String() instead of string(buf.Bytes())
- Add context.WithTimeout to compile and test commands to prevent hangs
- Set cmd.WaitDelay to force-close pipes from orphaned child processes

### DIFF
--- a/imports/go.mod
+++ b/imports/go.mod
@@ -1,3 +1,3 @@
 module github.com/tinygo-org/tinygo-site/imports
 
-go 1.16
+go 1.25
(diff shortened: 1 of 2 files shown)


aykevl (approved):
LGTM

deadprogram:
Thanks for review @aykevl now merging.

PR #487, opened 2026-04-18
boards: add SVG and improved layout for featured boards

deadprogram:
Adds some visual appeal:

image


aykevl:
Might be worth adding some lighter background behind them on the dark mode page so they become more visible, such as a white drop shadow?

PR #503, opened 2026-05-27
change xiao color

### LINKED ISSUE #502: Xiao SVG files
We need to update Xiao SVG files! They are NOT green. @deadprogram I think ellie sent us a couple SVGs :)

https://tinygo.org/docs/reference/microcontrollers/featured/

soypat:
Solves https://github.com/tinygo-org/tinygo-site/issues/502


deadprogram:
Thank you very much for fixing this @soypat now merging.

PR #504, opened 2026-05-29
feature: use updated XIAO boards and updates playground version with XIAO ESP32-S3

deadprogram:
This PR is to use updated XIAO boards and updates playground version with XIAO ESP32-S3


conejoninja:
Same issue as with the playground, the C3 is missing in the board list ¿?

deadprogram:
> Same issue as with the playground, the C3 is missing in the board list ¿?

That is because it does not have a built-in LED.

deadprogram:
Any other feedback from anyone before merge?

conejoninja:
Merged, thanks for your work.

PR #505, opened 2026-06-02
Add PWM peripheral documentation to reference/machine

cibomahto:
This adds documentation for the PWM interface, which is supported by many of the hardware platforms. I based these definitions off of an average of all of the implementations.


cibomahto:
Also see: https://github.com/tinygo-org/tinygo/issues/5441 for a list of the inconsistencies between implementations.

deadprogram:
Thanks for the addition here @cibomahto now merging.

PR #506, opened 2026-06-02
fix WASI example code and build command

jwetzell:


deadprogram:
Does the example/command not work? They are supposed to be equivalent to the same thing in Big Go.

jwetzell:
@deadprogram turned too many knobs at once I took the example from the [wazero link](https://wazero.io/languages/tinygo/) but it was really just the `//export add` that was needed. Corrected

aykevl (changes requested):
`//export` is an old pragma that should not be used anymore. Instead, you'll probably want to use `//go:wasmexport` instead.

jwetzell:
@aykevl I would have used that if it actually worked, but it does not. Using `//go:wasmexport` results in `panic: runtime error: //go:wasmexport function called before runtime initialization`.

aykevl:
@jwetzell that means you need to call `_initialize` on the module first. (We could improve the error message for this). This is not a bug, this is expected behavior (`//export` will also fail eventually if the runtime isn't initialized, but it won't tell you why it will just crash with a weird error if you write anything but trivial code).

jwetzell:
@aykevl I was simply testing by building the same example with both `//export` and `//go:wasmexport` and running them with `wasmtime`

```
package main

//export add
func add(x, y uint32) uint32 {
	return x + y
}

// main is required for the `wasip1` target, even if it isn't used.
func main() {}
```
VS
```
package main

//go:wasmexport add
func add(x, y uint32) uint32 {
	return x + y
}

// main is required for the `wasip1` target, even if it isn't used.
func main() {}
```
Then running them with `wasmtime run --invoke add main.wasm 1 1` the `//export` version results in the expected output of 2 while the `//go:wasmexport` has called before initialization. Seems like my testing was just wrong to assume that they would work the same way. 

I'll leave fixing this example to someone who can provide a full example.

jwetzell:
Even the most minimal test using wazero like below works just fine with `//export add` but fails with `//go:wasmexport add` with a stack trace to the same `//go:wasmexport function called before runtime initialization`. I'm admittedly not to well versed in WASM/WASI things so can't really dig into what the differences between `export` and `go:wasmexport` are but they are not interchangeable and I've only been able to get the former to work.

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/tetratelabs/wazero"
	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
)

func main() {
	ctx := context.Background()

	wasmBytes, err := os.ReadFile("main.wasm")
	if err != nil {
		panic(err)
	}

	r := wazero.NewRuntime(ctx)
	defer r.Close(ctx)

	wasi_snapshot_preview1.MustInstantiate(ctx, r)

	mod, err := r.InstantiateWithConfig(ctx, wasmBytes, wazero.NewModuleConfig().WithStartFunctions("_initialize"))
	if err != nil {
		panic(err)
	}
	res, err := mod.ExportedFunction("add").Call(ctx, 1, 2)
	if err != nil {
		panic(err)
	}
	fmt.Println(res[0])
}
```

aykevl:
I don't know why the wazero example doesn't work, it looks correct to me. Still, while a simple add function works with `//export` anything actually complex will crash since the runtime isn't initialized.

jwetzell:
I too don't know why it isn't working and I wasn't going to put something that I couldn't get working into an example on the site. The existing example is just completely wrong though so I guess no one is looking at it anyways.

PR #507, opened 2026-06-19
getting-started: Improve Arch linux getting-started install

AndersonQ:
It makes the Arch Linux package explicit by making the name of the package the link to the package. What also make both packages, the Arch package and the AUR package consistent.

### Note for the reviewer
It's my first PR here. I didn't find any specific guideline, so please let me know if there is a guideline to follow.


aykevl:
Thanks! Looks reasonable to me.

PR #511, opened 2026-09-02
docs: add a guide for the uefi-amd64 target

deadprogram:
TinyGo 0.42.0 added the `uefi-amd64` target, but the site has no documentation for it. This adds a guide.

### Changes

- `content/docs/guides/uefi.md`: a new guide with these sections.
  - How to write a first program with the `device/uefi` package.
  - How to build the program, and what the output file is.
  - How to run the program in QEMU with the OVMF firmware.
  - How to start the program from a USB drive on real hardware.
  - How to use the UEFI system table, the boot services, the console, events, and the unwrapped protocols.
  - The limitations of the target.
- `content/docs/reference/usage/important-options.md`: add `-target=uefi-amd64` to the list of example targets, with a link to the new guide.

### Notes

All three code examples in the guide compile with TinyGo 0.42.0 and give a PE32+ EFI application. I could not run QEMU on my machine, so the QEMU command comes from the reference below.

Reference: https://bsky.app/profile/af9.us/post/3muicbgfdvk2l

### DIFF
--- a/content/docs/guides/uefi.md
+++ b/content/docs/guides/uefi.md
@@ -1,0 +1,149 @@
+---
+title: "UEFI applications"
+weight: 2
+description: |
+  How to use TinyGo to create UEFI applications that the firmware starts before an operating system.
+---
+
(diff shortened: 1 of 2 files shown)


deadprogram:
Merging because we need it. Can edit afterward if needed.

PR #516, opened 2026-09-14
Fix note about WiFi support on ESP32 Xtensa LX6 based controllers.

remyzandwijk:
As of TinyGo 0.42, WiFi is supported on Xtensa LX6 processor based ESP32 microcontrollers.


deadprogram (approved):
Thank you!

